PR #18, opened 2015-08-19
Deletes instance objects in order so that no object keeps tangling in…

om-sharma:
Deletes instance objects in order so that no object keeps tangling in case we re-run the script


delete-instance.sh
```diff
@@ -3,7 +3,7 @@
 set -eu
 
 usage() {
-    echo "Usage: ./destroy-env.sh instance-name"
+    echo "Usage: $0 instance-name"
```
philandstuff: :+1:

PR #24, opened 2015-08-26
Tidy up IAM code

philandstuff:
This does quite a lot of stuff -- probably easier to read individual commit messages


create-code-deploy-application.sh
```diff
@@ -2,18 +2,9 @@
...
-
-fetchInstanceProfile(){
-    CODE_DEPLOYER_PROFILE_ARN=$(aws iam get-instance-profile \
-                                --instance-profile-name $INSTANCE_PROFILE_NAME \
-                                --query InstanceProfile.Roles[0].Arn \
-                                --output text)
-
-    echo "$CODE_DEPLOYER_PROFILE_ARN"
-}
+CODE_DEPLOYER_ROLE_ARN=arn:aws:iam::022990953738:role/code-deploy-role
```
om-sharma: can we fetch ARN by the role name instead of hardcoding it?

create-instance.sh
```diff
@@ -54,6 +54,72 @@ set_up_security_group() {
 
 }
 
+create_instance_profile() {
```
om-sharma: it is good to have a separate script to create instance profile, we can call the script here to create profile.

create-code-deploy-application.sh
philandstuff: discussed IRL. the ARN is entirely based on the role name so this isn't a big deal.

create-instance.sh
philandstuff: yup, will do

philandstuff:
@om-sharma updated

PR #25, opened 2015-08-27
Added create and delete scripts for indexer instance.

om-sharma:
Added create and delete scripts for indexer instance.
user-data.yaml doesn't need to setup postgres, kafka and zookeeper now.

PR #26, opened 2015-08-27
Authorizing new instance to access presentation rds db also.

om-sharma:
Authorizing new instance to access presentation rds db also. for now we are not asking the rds-instance name parameter when you run create instance script.

### DIFF
--- a/create-instance.sh
+++ b/create-instance.sh
@@ -3,12 +3,34 @@
 set -eu
 
 usage() {
-    echo "Usage: $0 <environment-name> <rds-instance-name>"
+    echo "Usage: $0 <environment-name>"
     echo ""
     echo "* Creates an EC2 instance, security group and route 53 entry for environment-name."
     echo "* Attaches IAM instance profile to the instance."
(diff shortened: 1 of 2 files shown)


create-instance.sh
```diff
@@ -126,41 +142,19 @@ EOF
...
-
 # ensure aws CLI is set up with needed profiles before continuing
 check_aws_profiles_exist "$DNS_PROFILES"
 
 ./create-appserver-instance-profile.sh "$INSTANCE_PROFILE_NAME"
 
-set_up_security_group "$SG" "$ENV" "$RESTRICTED_PORTS" "$PUBLIC_PORTS" "$DB_SG"
+set_up_security_group "$RESTRICTED_PORTS" "$PUBLIC_PORTS"
 
 USER_DATA=$(sed -e "s/%PGPASSWD%/${PG_PASSWORD}/" "${USER_DATA_FILE}" | base64)
```
philandstuff: fyi we can remove the PGPASSWD stuff when #25 is merged

PR #36, opened 2015-10-08
make sure that a new tag 'AppServer' is used to deploy

om-sharma:
Changes to make sure that a new tag 'AppServer' is used to deploy in specific environment


philandstuff:
can you update the `create-indexer-instance` script too?

om-sharma:
done

PR #37, opened 2015-10-08
fix role access permissions

philandstuff:
The ROLE_NAME is no longer the same as the REGISTER_NAME -- ROLE_NAME is
something like prod-school, but REGISTER_NAME is something like school,
which is what we need in the role access policy.


create-appserver-instance-profile.sh
```diff
@@ -15,7 +15,8 @@ fi
```
om-sharma: if condition must check equality for 3 parameters now.

philandstuff: good point

PR #43, opened 2015-10-27
Added script to launch register in one command

om-sharma:
Check commits


create-register.sh
```diff
@@ -0,0 +1,263 @@
+#!/bin/sh
+
+
+
+set -x
```
philandstuff: don't need this

create-register.sh
```diff
@@ -0,0 +1,263 @@
...
+REGISTER_DB_NAME=${REGISTER_NAME//-/_}
+CONFIG_FILE_DIRECTORY=${CURRENT_WORKING_DIR}/register-configurations
+
+function cdIntoCredentailsRepository(){
+  pushd ~/.registers-pass/
+  export PASSWORD_STORE_DIR=$(cd $(dirname -- $BASH_SOURCE) && pwd)
+}
+
+function createDatabaseAndUsers(){
+  cdIntoCredentailsRepository
```
philandstuff: not sure I like using `pushd` and `popd` in different scopes -- also have you considered using `cd` inside a subshell? eg:

```
(cd $dir
   foo
   bar
   baz
)
# back in original dir here
```

create-register.sh
```diff
@@ -0,0 +1,263 @@
...
+ENV=$2
+BUCKET_NAME=openregister.$ENV.config
+INDEXER_CONFIG_LOCATION=/tmp/indexer.properties
+CURRENT_WORKING_DIR=`pwd`
+
+REGISTER_DB_NAME=${REGISTER_NAME//-/_}
+CONFIG_FILE_DIRECTORY=${CURRENT_WORKING_DIR}/register-configurations
+
+function cdIntoCredentailsRepository(){
+  pushd ~/.registers-pass/
```
philandstuff: shouldn't this be `pushd $PASSWORD_STORE_DIR`?

create-register.sh
```diff
@@ -0,0 +1,263 @@
...
+  export PASSWORD_STORE_DIR=$(cd $(dirname -- $BASH_SOURCE) && pwd)
+}
+
+function createDatabaseAndUsers(){
+  cdIntoCredentailsRepository
+
+  BRANCH_NAME=adding_password_for_${ENV}_${REGISTER_NAME}
+
+  pass git checkout -b $BRANCH_NAME
```
philandstuff: is it worth explicitly creating a branch from `master` here, rather than whatever branch you happened to be on previously?

create-register.sh
```diff
@@ -0,0 +1,263 @@
...
+          PRESENTATION_USER_PASSWORD=`pass prod/rds/presentation/${REGISTER_DB_NAME}`
+          READ_ONLY_PRESENTATION_USER_PASSWORD=`pass prod/rds/presentation/${REGISTER_DB_NAME}_readonly`
+          ;;
+      *)
+          echo "Unrecognized environment: $ENV"
+          usage; exit 1
+          ;;
+  esac
+
+  CONFIG_FILE_DIRECTORY=register-configurations
```
philandstuff: duplicated from top of file?

create-register.sh
```diff
@@ -0,0 +1,263 @@
...
+BUCKET_NAME=openregister.$ENV.config
+INDEXER_CONFIG_LOCATION=/tmp/indexer.properties
+CURRENT_WORKING_DIR=`pwd`
+
+REGISTER_DB_NAME=${REGISTER_NAME//-/_}
+CONFIG_FILE_DIRECTORY=${CURRENT_WORKING_DIR}/register-configurations
+
+function cdIntoCredentailsRepository(){
+  pushd ~/.registers-pass/
+  export PASSWORD_STORE_DIR=$(cd $(dirname -- $BASH_SOURCE) && pwd)
```
philandstuff: this won't work because this bash file isn't in the password store dir, it's in the `deployment` dir...

om-sharma:
closing the pull request because the script was not tested. It is waiting for other pull request in credentials repository.

PR #45, opened 2015-11-02
added script to launch a register end-to-end

om-sharma:
Script to launch a register. 

The script generates the db user passwords and creates the database. 
After this it creates the indexer, mint and presentation configuration required for the register.
Next step, it syncs the configuration with the configurations in S3 buckets.
Creates the register instance and configure it.
In next step script deploys the mint app by choosing the last deployed mint artifacts from master repository. also grants access to readonly mint user.
If data file is specified it loads the data into mint api.
Deploys the indexer app and presentation app.

After the script execution is complete you can see the register at http://<register>.<env>.openregister.org

We had successfully launched many registers last week by using this script.

### DIFF
--- a/create-register.sh
+++ b/create-register.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
 
@@ -28,14 +28,12 @@ CURRENT_WORKING_DIR=`pwd`
 
 CONFIG_FILE_DIRECTORY=${CURRENT_WORKING_DIR}/register-configurations
 
-LOCAL_PASSWORD_STORE_DIR=~/.registers-pass
-
 INDEXER_CONFIG_LOCATION=/tmp/indexer.properties
 
-export PASSWORD_STORE_DIR=$LOCAL_PASSWORD_STORE_DIR
+export PASSWORD_STORE_DIR=~/.registers-pass
 
 function createDatabaseAndUsers(){
-  cd $LOCAL_PASSWORD_STORE_DIR
+  cd $PASSWORD_STORE_DIR
 
   BRANCH_NAME=adding_password_for_${ENV}_${REGISTER_NAME}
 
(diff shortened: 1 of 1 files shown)


create-register.sh
```diff
@@ -0,0 +1,257 @@
+#!/bin/sh
```
philandstuff: should probably be `/bin/bash` -- I suspect you'll have used some bash-specific features

create-register.sh
```diff
@@ -0,0 +1,257 @@
...
+
+CURRENT_WORKING_DIR=`pwd`
+
+CONFIG_FILE_DIRECTORY=${CURRENT_WORKING_DIR}/register-configurations
+
+LOCAL_PASSWORD_STORE_DIR=~/.registers-pass
+
+INDEXER_CONFIG_LOCATION=/tmp/indexer.properties
+
+export PASSWORD_STORE_DIR=$LOCAL_PASSWORD_STORE_DIR
```
philandstuff: why two separate variables?

create-register.sh
```diff
@@ -0,0 +1,257 @@
...
+    exit 1
+  fi
+
+  echo "$REGISTER_NAME.source.postgres.db.connectionString=jdbc:postgresql://$RDS_MINT_END_POINT:5432/${REGISTER_DB_NAME}?user=${REGISTER_DB_NAME}_readonly&password=${READ_ONLY_MINT_USER_PASSWORD}" >> $INDEXER_CONFIG_LOCATION
+  echo "$REGISTER_NAME.destination.postgres.db.connectionString=jdbc:postgresql://$RDS_PRESENTATION_END_POINT:5432/${REGISTER_DB_NAME}?user=${REGISTER_DB_NAME}&password=${PRESENTATION_USER_PASSWORD}" >> $INDEXER_CONFIG_LOCATION
+}
+
+function loadConfigurationToS3(){
+  cd $CURRENT_WORKING_DIR
+  aws s3 sync $CONFIG_FILE_DIRECTORY s3://$BUCKET_NAME
```
philandstuff: will this download the whole bucket? we should probably just copy what we need up rather than leaving unencrypted passwords on local machine

philandstuff: see also `aws s3 cp --recursive`

om-sharma: s3 sync doesn't download the whole bucket. it copies new and updated files from source to destination.

PR #47, opened 2015-12-11
Repeatable creation of etcd cluster

AndrewGorton:
Uses Terraform for repeatable provisioning of a CoreOS ETCD cluster.

Terraform also creates the VPC, internet gateway, subnet, associated routing table, security group and EC2 instances.


AndrewGorton:
ETCD broken.

PR #48, opened 2015-12-11
Repeatable creation of etcd cluster

AndrewGorton:
Repeatable creation of an ETCD cluster in AWS.

Uses [Terraform](http://www.terraform.io) to create VPC, subnet, internet gateway, routing gateways, security groups and EC2 instances.

Uses the public discovery.etcd.io service to get a cluster identifier (and for cluster members to locate each other).


terraform/instances.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_instance" "coreos-tf-example" {
```
om-sharma: can we remove `-example` from the instance name to make it consistent with other resources.

AndrewGorton: Have removed it and renamed to match other resource identifiers.

PR #65, opened 2016-02-23
Added inline policy to so Indexer to publish metrics to CloudWatch.

AndrewGorton:


AndrewGorton:
@slawekm Need to add this to Terraform too.

PR #66, opened 2016-02-26
Add config and IAM access to allow Mint to heartbeat to CloudWatch.

AndrewGorton:
This needs merging after https://github.com/openregister/mint/pull/65, otherwise Mint will complain of an unknown configuration directive.

The code which reads the configuration is safe - if the config isn't set, then heartbeating is not enabled.


philandstuff:
> This needs merging before openregister/mint#65, otherwise Mint will complain of an unknown configuration directive.

do you mean "after"?

AndrewGorton:
Yep - corrected! (oops)

PR #91, opened 2016-05-24
add school register in alpha environment

om-sharma:


aws/registers/register_school.tf
```diff
@@ -0,0 +1,57 @@
+module "school_policy" {
+  source = "../modules/instance_policy"
+  id = "school"
+  enabled = "${signum(lookup(var.instance_count, "school"))}"
```
philandstuff: do we need to set this variable somewhere?

om-sharma: `instance_count` for every environment comes from `<environment>.tfvars` file, in this case `alpha.tfvars`.

PR #93, opened 2016-06-21
bump to latest, minimal, amazon linux AMI

philandstuff:
This bumps us to the latest AMI for the latest Amazon linux release -
2016.03.2.  It also moves us to the "minimal" ami which is smaller and
comes with less stuff preinstalled.

As a result, we need to install "aws-cli" and "ruby" packages to ensure
the codedeploy agent setup code works.

I've tested this on the test register register presentation box.

### DIFF
--- a/aws/modules/indexer/variables.tf
+++ b/aws/modules/indexer/variables.tf
@@ -7,7 +7,7 @@ variable "cidr_block" {}
 variable "db_cidr_block" {}
 
 variable "instance_ami" {
-  default = "ami-a10897d6"
+  default = "ami-6ced7f1f"
 }
 
 variable "instance_count" {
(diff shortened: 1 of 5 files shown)


michaelabenyohai:
Do we care that we haven't also updated the old scripts to use ami-6ced7f1f?

create-indexer-instance.sh
create-instance.sh

And that we have not added the new packages to old user-data.yaml (though ssh keys are also out of date in this file)?

philandstuff:
yes I think that's fine. Because I moved to the minimal iso, it isn't simply a drop-in replacement; you have to ensure you install ruby and aws-cli as well. Once we are rid of the preview environment we don't need those scripts anymore and they can be deleted.

PR #95, opened 2016-06-22
automatic updates

philandstuff:
install yum-cron to handle automatic updates.

Also, enable the docker service so that it restarts after reboot.


philandstuff:
oops, the original version of this PR didn't actually install `yum-cron`...

PR #101, opened 2016-07-06
Create discovery environment

michaelabenyohai:
With the following registers:
register
datatype
field
local-authority
local-authority-type

### DIFF
--- a/aws/modules/load_balancer/main.tf
+++ b/aws/modules/load_balancer/main.tf
@@ -1,7 +1,7 @@
 resource "aws_elb" "load_balancer" {
   count = "${var.enabled}"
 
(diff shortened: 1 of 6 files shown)


aws/modules/load_balancer/main.tf
```diff
@@ -1,7 +1,7 @@
 resource "aws_elb" "load_balancer" {
   count = "${var.enabled}"
 
-  name = "${replace(format("%s-%s-elb", var.vpc_name, var.id),"_","-")}"
+  name = "${replace(format("%s-%s-elb", replace(var.vpc_name,"discovery","disco"), var.id),"_","-")}"
```
philandstuff: worth commenting why this substitution is made?

PR #106, opened 2016-07-29
Add openregister configs

chr1sdi:


ansible/upload_configs_to_s3.yml
```diff
@@ -10,9 +10,11 @@
     services:
       mint: 'mint-config.yaml'
       presentation: 'config.yaml'
+      openregister: 'openregister-config.yaml'
```
philandstuff: this should just be `config.yaml`

ansible/create_databases.yml
```diff
@@ -24,3 +24,11 @@
         write_user: 'mint',
         master_password: "{{ lookup('pass', '{{ vpc }}/rds/mint/master') }}"
       }
+
+    - { role: database,
+        inventory_group_name: '{{ vpc }}-openregister-db',
+        read_user: 'openregister',
```
philandstuff: this needs more thought

chr1sdi: it should only add permission to an existing user:
http://docs.ansible.com/ansible/postgresql_user_module.html#synopsis

the order of creating users is WRITE, READ so the second call (privilege `SELECT`) shouldn't do anything as WRITE has `ALL`

ansible/roles/service_config/templates/openregister-config.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+  initialSize: {{ db_initial_size | default(1) }}
+  minSize: {{ db_min_size | default(1) }}
+  maxSize: {{ db_max_size | default(4) }}
+
+  properties:
+    charSet: UTF-8
+
+registerDomain: {{ register_domain }}
+
+enableDownloadResource: {{ settings.enable_download_resource | default(true) }}
```
philandstuff: this isn't a complete, valid config file for openregister-java. In particular, it's missing the `register:` field.

PR #108, opened 2016-08-18
Deploy merged openregister

michaelabenyohai:
Replace mint and presentation instances with openregister instances.


ansible/create_databases.yml
```diff
@@ -2,7 +2,8 @@
 - name: Create databases
   hosts:
     - '&tag_Environment_{{ vpc | normalize_ec2_tags }}'
-    - '&tag_Role_indexer_app'
+    - '&tag_Register_territory' # FIXME
+    - '&tag_Role_openregister_app' # FIXME
```
philandstuff: hmm, do we need to talk about fixing this? Previously we used `indexer` as it had two nice properties:

 - it had access to both RDS instances
 - it was guaranteed to exist in each environment

I'm happy to merge this PR as is to get it in master, but a brief discussion about how we iron out this wrinkle might be good.

ansible/roles/terraform_config/templates/register_template.tf.j2
```diff
@@ -41,6 +41,23 @@ module "{{ item }}_mint" {
   user_data = "${template_file.user_data.rendered}"
 }
 
+module "{{ item }}_openregister" {
```
philandstuff: we should delete mint and indexer from this template

aws/modules/instance_policy/iam_policy_config_access.tf
```diff
@@ -12,6 +12,7 @@ resource "aws_iam_role_policy" "policy_config_access" {
       ],
       "Resource": [
         "arn:aws:s3:::openregister.${var.vpc_name}.config/${var.id}/mint/*",
+        "arn:aws:s3:::openregister.${var.vpc_name}.config/${var.id}/openregister/*",
         "arn:aws:s3:::openregister.${var.vpc_name}.config/${var.id}/presentation/*",
```
philandstuff: ditto

PR #110, opened 2016-08-30
Migrate register instances to openregister

michaelabenyohai:
These are the last few EC2 instances to migrate to openregister-java or switch off.

Address mint/presentation in discovery/test have not yet been destroyed just in case loading the new address register in test proves difficult. Once this is done, these can be destroyed - they are the last mint/presentation EC2 instances left.

Mint/presentation RDS instances have not been destroyed yet for the same reason but can be done as a separate task later on.


philandstuff:
👍

PR #114, opened 2016-09-02
Remove new-registers.yml and new-fields.yml

michaelabenyohai:


philandstuff:
I'll wait merging this until we can confirm that nothing is using the new-fields.yaml anymore.

michaelabenyohai:
@philandstuff should be possible to merge and apply this now. I believe there is now nothing using these config files.

PR #118, opened 2016-09-08
Correct typo so that Chris is added to the users group

michaelabenyohai:
This needs to be applied once merged.


michaelabenyohai:
This needs to be applied once merged.

PR #129, opened 2016-09-12
Remove mint, indexer and presentation terraform modules

michaelabenyohai:
Also remove mint, indexer and presentation from ansible tasks and templates.

Needs to be applied once merged.


ansible/roles/terraform_config/templates/environment.tfvars.j2
```diff
@@ -12,38 +12,24 @@ admin_ips = "{{ admin_ips }}"
...
-{% set presentation_database_cidr_block_default_b = vpc_cidr_block | ipsubnet(27, 161) -%}
-{% set presentation_database_cidr_block_default_c = vpc_cidr_block | ipsubnet(27, 162) -%}
-{% set presentation_database_cidr_block_defaults = [
-  presentation_database_cidr_block_default_a, 
-  presentation_database_cidr_block_default_b, 
-  presentation_database_cidr_block_default_c 
+openregister_cidr_block = "{{ openregister_cidr_block | default(openregister_cidr_block_defaults) }}"
+{% set openregister_database_cidr_block_default_a = vpc_cidr_block | ipsubnet(27, 160) -%}
+{% set openregister_database_cidr_block_default_b = vpc_cidr_block | ipsubnet(27, 161) -%}
+{% set openregister_database_cidr_block_default_c = vpc_cidr_block | ipsubnet(27, 162) -%}
```
michaelabenyohai: @philandstuff please can we talk about whether these defaults are still sensible for new openregister app.

philandstuff: yeah, we should make all subnets a `/24` at least to avoid issues of running out of IPs like we did with the public cidr blocks before.

PR #139, opened 2016-09-30
Change ansible scripts to configure register history urls

michaelabenyohai:
Add register history URLs as variables to thetag_Environment_xxx files. If the variable does not exist for the register, do not include in the template for creating config.yaml.


ansible/group_vars/tag_Environment_alpha
```diff
@@ -21,3 +21,7 @@ register_domain: alpha.openregister.org
 register_settings:
```
philandstuff: oh cool, I didn't know about this. nice to reuse it

PR #158, opened 2016-11-01
Use AWS Managed NAT Gateway

samcrang:
Deploying this change will mean instances in private subnets will use the [AWS Managed NAT Gateway](https://aws.amazon.com/blogs/aws/new-managed-nat-network-address-translation-gateway-for-aws/) to route traffic outbound.

We still need to tidy up the exisiting NAT instances so that they doen't have two roles (Bastion and NAT Gateway), but it seems sensible to do these as two separate deploys.

### DIFF
--- a/aws/modules/core/gateways.tf
+++ b/aws/modules/core/gateways.tf
@@ -6,10 +6,3 @@ resource "aws_internet_gateway" "igw" {
     Environment = "${var.vpc_name}"
   }
 }
-
-/*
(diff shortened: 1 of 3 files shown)


samcrang:
I've changed this to explicitly only add a NAT gateway to the first public subnet (this is essentially the same behaviour as currently exists) in order to avoid having to change more things in the same deploy.

PR #160, opened 2016-11-04
Getting only first host that matches filter

karlbaker02:


philandstuff:
doesn't seem to work for me 😕

philandstuff:
I'm testing by replacing the last task with a debug one showing which hosts matched:

```yaml
- hosts: collected_hosts
  name: foo
  tasks:
    - debug: msg="System {{ inventory_hostname }} has uuid {{ ansible_product_uuid }}"  
```

PR #161, opened 2016-11-04
Pin terraform version

philandstuff:
This sets an expected terraform version in the Makefile and runs a check
to ensure that we are running the expected version of terraform.  This
is to avoid inadvertantly running a newer version of terraform and thus
migrating the statefile to a non-backward-compatible version.

### DIFF
--- a/aws/registers/Makefile
+++ b/aws/registers/Makefile
@@ -29,7 +29,6 @@ module_depth=-1
 check-for-local-state: 
 	@test ! -f terraform.tfstate || exit 1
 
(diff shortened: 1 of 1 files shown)


philandstuff:
hmm, having upgraded to 0.7.8, I now get this error message:

    $ make plan -e vpc=alpha
Makefile:19: *** Expected terraform version v0.7.8, but saw v0.7.8.  Stop.

😞

philandstuff:
pushed fix for that error ☝️

PR #162, opened 2016-11-04
Fix bug where the same data got minted multiple times

philandstuff:
The mint_data.yml playbook had a problem when minting to a register
which had multiple app servers: it would run the minting task once per
appserver, rather than once.  This meant that you would get as many
copies of data as you had appservers.

This fixes the problem by selecting the first appserver in the
`collected_hosts` group, rather than the whole group.


karlbaker02:
👍

PR #163, opened 2016-11-04
Generating new directory with random name for keeping minting data

karlbaker02:


ansible/roles/service_data/tasks/data.yml
```diff
@@ -1,6 +1,13 @@
 - name: Ensure target directory exists
   file: state=directory dest="{{ workdir }}"
```
philandstuff: do we still need this task?

ansible/roles/service_data/tasks/data.yml
```diff
@@ -1,6 +1,13 @@
...
 
+- name: Ensure target directory exists
+  command: mktemp -d /tmp/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
philandstuff: why specify our own template? Isn't the default good enough?

philandstuff:
oh, and a minor style thing: I prefer [tim pope's git message style](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), in particular:

> Write your commit message in the imperative: "Fix bug" and not "Fixed bug"
or "Fixes bug."

to which I would add: not "Fixing bug" either.

PR #166, opened 2016-11-08
remove unused variable

philandstuff:
unused since #163.


philandstuff:
@karlbaker02 does this make sense to you?

PR #167, opened 2016-11-08
Simplify ansible host selection in mint_data.yml

philandstuff:
This adds our shiny new route53 private hosted zones to our ansible
inventory.  This means we have groups for each instance corresponding to
the route53 record name.  For example, there's a group called
`country-1.test.openregister` which contains one host, the first
appserver for the country register in the test vpc.

We can now use these names to more easily identify a single appserver
for a register within a VPC in mint_data.yaml.

### DIFF
--- a/ansible/ec2.ini
+++ b/ansible/ec2.ini
@@ -39,7 +39,7 @@ vpc_destination_variable = private_ip_address
 
 # To tag instances on EC2 with the resource records that point to them from
 # Route53, uncomment and set 'route53' to True.
-route53 = False
+route53 = True
 
 # To exclude RDS instances from the inventory, uncomment and set to False.
 rds = True
--- a/ansible/mint_data.yml
+++ b/ansible/mint_data.yml
@@ -8,24 +8,8 @@
       when: registers is none
       always_run: yes
 
-- name: Compute mint hosts
-  hosts: localhost
-  vars:
-    vpc_tag: 'tag_Environment_{{ vpc }}'
-  tasks:
-    - name: query openregister hosts
-      set_fact:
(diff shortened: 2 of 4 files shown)


ansible/mint_data.yml
```diff
@@ -8,24 +8,8 @@
...
-
-    - name: include matching hosts
-      add_host: name="{{ item }}" group="collected_hosts"
-      with_items: "{{ openregister_hosts }}"
-      when: hostvars[item].ec2_tag_Register == registers
-      always_run: yes
-
 - name: Load data
-  hosts: "collected_hosts[0]"
+  hosts: "{{ registers }}-1.{{ vpc }}.openregister"
```
michaelabenyohai: What happens if the `registers` variable is a list? I believe this may have been broken in the previous commit as well.

The deployment README suggests that a list of register names can be passed. We should either fix this or change the variable name to `register` and update the README.

philandstuff: good point, will fix

philandstuff: @michaelabenyohai updated

ansible/ec2.ini
```diff
@@ -39,7 +39,7 @@ vpc_destination_variable = private_ip_address
 
 # To tag instances on EC2 with the resource records that point to them from
 # Route53, uncomment and set 'route53' to True.
-route53 = False
+route53 = True
```
philandstuff: hmm, it turns out I didn't need this after all. Might leave it in though, seems useful? 🤔

samcrang: 👍

PR #172, opened 2016-11-14
Redirect job-centre to jobcentre

philandstuff:
This adds a new module, `redirect`, which allows you to redirect all
requests from one domain to another.  It's built from an S3 static
website (which does the actual redirecting), a CloudFront
distribution (which is just there to support HTTPS) and a Route 53
record.

It also instantiates the module for the job-centre discovery register,
redirecting to the jobcentre register instead.


philandstuff:
I've deployed this but it takes ages for a cloudfront distribution to provision so it's not active yet.

PR #178, opened 2016-11-17
Send Cloudfront logs to Sumologic

samcrang:
This should be mostly self-explanatory. The Terraform here runs separately from any of the other Terraform code in this repo.

I've avoided working out how much of an IPv6 address to remove in order to "anonymize" it by setting IPv6 address (which we don't currently use anyway) to `-`.


aws/cloudfront/lambda/cloudfront-s3-logs-to-sumologic/README.md
```diff
@@ -0,0 +1,14 @@
+# Cloudfront to Sumologic Lamdbda function
```
philandstuff: typo

aws/cloudfront/main.tf
```diff
@@ -0,0 +1,76 @@
+provider "aws" {
+  region = "${var.aws_region}"
```
samcrang: Sounds legit. 👍

PR #179, opened 2016-11-18
Migrate registers to new module

philandstuff:
This migrates the rest of the registers to the new terraform module, created in #174.

I've already fully applied this to test and discovery, and have started applying this to beta, but I haven't touched alpha yet.


philandstuff:
This has now been deployed to all environments successfully.

aws/modules/register/elb.tf
```diff
@@ -2,7 +2,7 @@ resource "aws_elb" "load_balancer" {
   count = "${signum(var.instance_count)}"
 
   # replace discovery with disco to ensure name does not exceed string length limit
-  name = "${replace(format("%s-%s", replace(var.vpc_name,"discovery","disco"), var.id),"_","-")}"
+  name = "${replace(format("%s--%s", replace(var.vpc_name,"discovery","disco"), var.id),"_","-")}"
```
michaelabenyohai: Is there a reason for having the extra dash? Is it because you rebuilt the world whilst always having a load balancer running?

PR #196, opened 2016-12-13
Add queues and triggers for notifications from CodeDeploy

samcrang:
We need this so we can notify to our Slack channel every time we deploy. Currently we're still manually deploying the Lambda function that hits the Slack Webhook as well as the triggers to glue SNS and Lambda together.


karlbaker02:
👍

PR #198, opened 2016-12-14
Use Terraform for CloudFront configuration

samcrang:
I've mirrored the configuration of what has been manually deployed, except for the TTLs which I've kept low until we know it's all good. I also specifically haven't changed DNS on *.register.gov.uk for the same reason.

### DIFF
--- a/aws/modules/register/cdn.tf
+++ b/aws/modules/register/cdn.tf
@@ -1,0 +1,58 @@
+resource "aws_cloudfront_distribution" "cloudfront" {
+  count = "${signum(var.enable_cdn * var.instance_count)}"
+  enabled = true
+
+  aliases = ["${var.id}.register.gov.uk"]
+
+  http_version = "http1.1"
+
+  origin {
+    domain_name = "${aws_route53_record.load_balancer.fqdn}"
+    origin_id   = "register_origin"
+
+    custom_origin_config {
+      http_port = 80 # "required" but not used
+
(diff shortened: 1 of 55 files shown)


samcrang:
I think there's some work that's needed to tidy up how `cloudfront_certificate_arn` is used—it's only actually used in discovery and beta but has values for all environments.

aws/modules/register/cdn.tf
```diff
@@ -0,0 +1,58 @@
...
+      https_port = 443
+      origin_protocol_policy = "https-only"
+      origin_ssl_protocols = ["TLSv1.2"]
+    }
+  }
+
+  default_cache_behavior {
+    allowed_methods = ["GET", "HEAD"]
+    cached_methods = ["GET", "HEAD"]
+    default_ttl = 300
```
philandstuff: we should talk about `Vary` headers at some point

aws/registers/register_company.tf
```diff
@@ -14,5 +14,6 @@ module "company" {
   dns_zone_id = "${module.core.dns_zone_id}"
   certificate_arn = "${var.elb_certificate_arn}"
 
+  enable_cdn = "${var.enable_cdn}"
```
philandstuff: I don't completely understand what happens if `enable_cdn` is true, but `cloudfront_certificate_arn` is unset

PR #199, opened 2016-12-14
Revert "Use Terraform for CloudFront configuration"

samcrang:
I'm backing out of these changes because it requires slightly more
effort that I'd originally hoped, specifically:

- We can't run more than one distribution that responds to the same host
(or "alias"). For example we can't run two separate distributions that
both contain aliases for "register.register.gov.uk".
- We want to refactor this so that the configuration of a CDN that
fronts a register is separate from the configuration that describes the
rest of the register infrastructure.

This reverts commit f050b9f5e7cc2aaff11d636cc65ea13384aec5d9.


philandstuff:
🔙

PR #204, opened 2016-12-16
Remove unrequired registers from discovery.

johnollier:


karlbaker02:
You might want to double check whether or not we need to delete job-centre (as we're dropping the database for it)

samcrang:
@karlbaker02, @johnollier should be fine—it's been renamed `jobcentre` and the redirect happens in CloudFront so nothing behind it should need to know about `job-centre`

karlbaker02:
👍

PR #205, opened 2016-12-16
Remove destructive, infrequently used tasks

samcrang:
These tasks seem a bit scary. I feel like the need to remove configs,
buckets and state happens so rarely it's not worth the risk of an
accidental tab-completion mistake from trashing important stuff.


philandstuff:
👍

PR #206, opened 2016-12-16
Add `diff-config` task

samcrang:
It's currently clumsy having to manually check the variables you want to
push haven't been changed by someone else after you last pulled them.
Hopefully, this should make things slightly easier/safer.

I'm not overly happy with the `bash -c` bit but it means we can use
process substitution to avoid copying files locally.


philandstuff:
do we want to add this to README?

samcrang:
Yes we should

PR #207, opened 2016-12-16
Remove reference to explicit Terraform version in documentation

samcrang:
The Makefile referenced specifies the version of Terraform we're using.


philandstuff:
also, it's not possible to install a specific version using homebrew 😢

PR #208, opened 2016-12-16
Undo work to automate CloudFront distributions

samcrang:
It turns out that doing this migration without any downtime is more
involved/fiddly that we'd hoped.

We had originally planned to temporarily run the beta registers without
any CDN so that we could migrate from the manually configured
distributions to the new Terraform managed distributions. Running
without the CDN is only required because CloudFront does not allow you
to run multiple, separate distributions that respond to the same host.
However, it is not possible to run `*.register.gov.uk` registers without
CloudFront because this is where TLS termination happens.

Given that the multi-tenancy work is happening soon we've decided to
shelve this until after that because this will allow us to have one
CloudFront distribution and it'll be simpler to migrate.

😭


michaelabenyohai:
😭 

please can you remove the relevant variables from the config in S3 too

samcrang:
Done!

PR #211, opened 2016-12-22
Multi tenancy register

chr1sdi:
https://trello.com/c/VLrT7YOk/524-deploy-a-multitenanted-register


samcrang:
We paired on this (like bosses) so I'm mergin'.

PR #220, opened 2017-01-11
Add address to the multi-tenanted register group in discovery

samcrang:
Address is currently small in discovery so it seems sensible to run it next to the other registers.


philandstuff:
is the config.yaml also updated?

samcrang:
I don't want to generate and push it until it's Approved

PR #228, opened 2017-01-17
Remove dead ansible

philandstuff:
these scripts weren't used


ansible/generate_terraform_register.yml
```diff
@@ -1,18 +0,0 @@
----
-- name: Generate register manifest for terraform
```
samcrang: This still gets referenced here: https://github.com/openregister/deployment#terraform-config-for-a-register

philandstuff:
@samcrang fixed

PR #229, opened 2017-01-18
Require particular ansible version

philandstuff:
This pins ansible to 2.2.1.0.  This isn't yet available in homebrew (see Homebrew/homebrew-core#8992) but you can install it with `pip install ansible` (into a virtualenv, if you prefer).


samcrang:
```
$ ansible-playbook upload_configs_to_s3.yml -e vpc=test -e sync=false
# works fine

$ git checkout require-particular-ansible-version

$ ansible-playbook upload_configs_to_s3.yml -e vpc=test -e sync=false
ERROR! the role 'openregister_config' was not found in /Users/samcrang/src/registers/deployment/ansible/includes/roles:/Users/samcrang/src/registers/deployment/ansible/includes:/etc/ansible/roles

The error appears to have been in '/Users/samcrang/src/registers/deployment/ansible/upload_configs_to_s3.yml': line 14, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - openregister_config
      ^ here

$ ansible-playbook --version
ansible-playbook 2.1.0.0
  config file = /Users/samcrang/src/registers/deployment/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```

philandstuff:
huh 😕

samcrang:
I upgraded Ansible to the correct version and it magically started working. I changed the required version to something else manually just to see it fail. :shipit:

PR #232, opened 2017-01-19
[RFC] add .editorconfig

philandstuff:
This adds an [editorconfig file][1] to force newline-at-end-of-file.

This is automatically recognised by IntelliJ IDEA and other JetBrains
products.  There are plugins for other editors:

 - vim: https://github.com/editorconfig/editorconfig-vim#readme
 - emacs: https://github.com/editorconfig/editorconfig-emacs#readme
 - sublime:
 - https://github.com/sindresorhus/editorconfig-sublime#readme

How do people feel about this? 👍 / 👎 ?

[1]: http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+end_of_line = lf
+insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,5 @@
+root = true
+
 [*]
 end_of_line = lf
 insert_final_newline = true


philandstuff:
@karlbaker02 would you be able to test if this works for you?

karlbaker02:
@philandstuff I'll have a look at this later on

samcrang:
I think we'd want to set `root = true` in our `.editorconfig` here. Otherwise, I totally dig it.

PR #237, opened 2017-01-20
Add config items for fields and registers yaml defaulting to s3 url.

johnollier:


ansible/roles/openregister_config/templates/openregister-config.yaml.j2
```diff
@@ -10,6 +10,11 @@
   {%- set app_user = "openregister" -%}
   {%- set app_password = lookup("pass", "{{ vpc }}/app/mint/" + reg).decode('utf-8') -%}
   {%- set settings = register_settings[reg] | default({}) -%}
+  {%- set field_register_url = "http://field." + register_domain + "/records.yaml?page-size=1000"  -%}
+  {%- set register_register_url = "http://register." + register_domain + "/records.yaml?page-size=1000"  -%}
```
samcrang: even though this is just delaying the inevitable do we want to just make this the largest value for `page-size`?

chr1sdi:
👍

PR #238, opened 2017-01-20
Make register register read from yaml configuration

chr1sdi:


michaelabenyohai:
I'm not sure I understand why these settings are per-register and not per-register-group

PR #243, opened 2017-01-25
Remove data duplication

philandstuff:
The `registers` variable contains duplicate data to the
`register_groups` variable.  This means that when we add a new
register we need to remember to add it to both lists.

This commit changes the ansible scripts to use
`register_groups.values()`, flattened, in order to have a single
source of truth for which registers exist in an environment.

(This will also make the README simpler when I update it)


chr1sdi:
👍

PR #244, opened 2017-01-25
Set trim_blocks to false

karlbaker02:
This PR is to remove the hack we previously had to make where we inserted a new line after conditional template blocks. Failure to insert a new line after a conditional block would lead to the situation along the lines of `custodianName: JohnSmithenableDownloadResource: true`.

In the config template file I am specifying `trim_blocks: False` to force Jinja2 to keep the first new line after a template block; by default it will remove the first new line (see http://docs.ansible.com/ansible/template_module.html towards the bottom).


chr1sdi:
this creates some empty spaces between certain properties - this is planned to be cleaned up in the later story when the config file is going to be reviewed... so 👍

PR #247, opened 2017-01-30
Create separate dns record resources

samcrang:
I still need to `push-config`s and generate the address config in the discovery environment. I'll do this when/if we this gets merged.


samcrang:
There's also so tidying up we can do (specifically around our `.tfvars`) that can happen after we deploy.

ansible/group_vars/tag_Environment_discovery
```diff
@@ -20,8 +20,10 @@ register_groups:
     - datatype
     - field
 
-  multi:
+  address:
     - address
```
philandstuff: wait, why are you doing this? `address` only needs to be separate in `test`, right?

aws/modules/register/variables.tf
```diff
@@ -1,74 +1,9 @@
...
-}
-
-variable "dns_zone_id" {
-  description = "Route 53 zone id for DNS records for ELBs (ie public records)"
-}
-
-variable "dns_ttl" {
-  default = "300"
-  description = "TTL, in seconds, for Route 53 DNS records"
+variable name {}
```
philandstuff: why did you remove the `description`s?

philandstuff:
there's a typo in `availiabilty_checks.tf ` filename 😆

ansible/group_vars/tag_Environment_discovery
philandstuff: ah I see you explained in bc8313b9aaf57ae3bf97f9c0dcbd4fd71488d775

aws/modules/register_group/dns.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_route53_record" "multi" {
+  count = "${length(var.registers)}"
```
philandstuff: what's this?

aws/registers/registers.tf
```diff
@@ -0,0 +1,448 @@
...
+
+  name = "territory"
+  environment = "${var.vpc_name}"
+  load_balancer = "${module.multi.load_balancer}"
+  dns_zone_id = "${aws_route53_zone.temporary_hosted_zone.zone_id}"
+
+  enable_availability_checks = "${var.enable_availability_checks}"
+}
+
+module "uk_register" {
```
philandstuff: this will need an update to the README (probably after #246 is merged) to document an extra step for creating a register

aws/modules/register_group/dns.tf
samcrang: you're right--this needs to go.

philandstuff:
other than those comments ☝️, this looks good

aws/registers/registers.tf
samcrang: Yup, I was just about to add a comment on your PR about this 😉

aws/modules/register_group/dns.tf
samcrang: Actually this is still needed so we don't cause any downtime when we deploy these changes. This can be removed later as part of cleanup.

This PR creates a temporary hosted zone which contains the records deployed in the new way. I'll swap over the `NS` records on `openregister.org` manually to this new Hosted Zone before doing some cleanup, applying these DNS changes to the Hosted Zone that's created by the `core` Terraform module, and swapping the `NS` records back.

samcrang:
@philandstuff how's this?

PR #249, opened 2017-01-30
Begin cleanup of old-style DNS and availability checks

samcrang:
There's a bit of temporary oddness (see: the line with `count = "${var.enabled ? 2 : 0}"`) here which is required whilst we move between hosted zones.

We need to make sure we deploy DNS records created with the new Terraform resources with both the temporary and "proper" hosted zone (the one managed by the `core` module). It is packaged up as a single revertable commit which can be deployed after we deploy these changes.

Heads up: you'll need to be careful with `terraform apply` to make sure you don't target the `module.core.aws_route53_record.zone_delegation` resource which will, if applied, delegate NS records on `<env>.openregister.org` to a hosted zone with potentially inconsistent records.


aws/modules/register_group/dns.tf
```diff
@@ -1,11 +0,0 @@
...
-  count = "${length(var.registers)}"
-  zone_id = "${var.dns_zone_id}"
-  name = "${element(var.registers, count.index)}.${var.vpc_name}.${var.dns_domain}"
-  type = "A"
-  alias {
-    name = "${aws_elb.load_balancer.dns_name}"
-    zone_id = "${aws_elb.load_balancer.zone_id}"
-    evaluate_target_health = false // we don't have anything to fail over to at the moment
-  }
-}
```
philandstuff: can we remove the `enable_availability_checks` and `registers` variables from variables.tf here, too?

PR #250, opened 2017-01-31
Allow Terraform patch-level version updates

samcrang:
It's annoying to have to bump the expected version of Terraform we use
for every individual Terraform release.

We're going to allow patch-level Terraform auto-updates with the
assumption that upgrades inside the same minor release are unlikely to
break things in a way that can't easily be fixed.

We still need to update the old `v0.7.x` usages of Terraform but those 
changes will be more involved and can happen separately.

### DIFF
--- a/aws/registers/Makefile
+++ b/aws/registers/Makefile
@@ -13,11 +13,11 @@
 	diff-config \
 	push-state
 
-export expected_terraform_version=v0.8.0
-export actual_terraform_version=$(shell terraform version | head -n 1 | cut -f2 -d" ")
+export expected_terraform_version=v0.8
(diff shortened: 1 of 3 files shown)


aws/registers/Makefile
```diff
@@ -13,11 +13,11 @@
 	diff-config \
 	push-state
 
-export expected_terraform_version=v0.8.0
-export actual_terraform_version=$(shell terraform version | head -n 1 | cut -f2 -d" ")
+export expected_terraform_version=v0.8
+export actual_terraform_version=$(shell terraform version | head -n 1 | cut -f2 -d" " | cut -d"." -f1,2)
```
philandstuff: 👍 nice

PR #258, opened 2017-02-03
Prefer source/destination security groups over CIDRs

samcrang:
This is a refactoring to make all security group rules reference other security groups to restrict source/destination traffic. We are no longer using CIDRs (other than `0.0.0.0/0` and office IPs) to control traffic.

This hopefully captures the intention behind rules as well as making things slightly more secure(?).


samcrang:
@chr1sdi says it's too early on a Friday not to deploy.

PR #267, opened 2017-02-07
Enable confirmation servers for StatusCake tests

samcrang:
By default StatusCake uses 2 confirmation servers in order to trigger
downtime events. It's only been since Terraform 0.8.5 that this option
has been exposed. Prior to Terraform 0.8.5 this was set to 0 which
meant caused downtime events that only one StatusCake node saw.

This doesn't feel ideal as I'd like to see StatusCake be more reliable
but I think this should give us additional confidence that when alerts
fire our service is really down.


samcrang:
Requires #265 to be merged first

karlbaker02:
👍

PR #268, opened 2017-02-07
Set SOA negative cache TTL to one hour

samcrang:
Without this a `NXDOMAIN` response will be cached by recursors for 24
hours. Given we're hoping to add registers fairly regularly (which will
require new subdomains) it seems sensible to keep this short in order to
avoid the case where someone accidentally(?) tries to resolve a new
register before the records have been created.

The `name-server` field is being set to the first name server in the
list of name servers the hosted zone uses. This is likely to be
different from the value created by AWS, however, it should be
acceptable to set it to any name server that responds authoritatively
for the domain

See: http://www.zytrax.com/books/dns/ch8/soa.html

### DIFF
--- a/aws/modules/core/route53.tf
+++ b/aws/modules/core/route53.tf
@@ -18,6 +18,14 @@ resource "aws_route53_record" "zone_delegation" {
   records = [ "${aws_route53_zone.core.name_servers}" ]
 }
 
+resource "aws_route53_record" "soa" {
+  zone_id = "${aws_route53_zone.core.zone_id}"
+  name = "${var.vpc_name}.${var.dns_domain}"
+  type = "SOA"
+  ttl = "900"
+  records = [ "${aws_route53_zone.core.name_servers.0}. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 ${var.soa_negative_cache_ttl}" ]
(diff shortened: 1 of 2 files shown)


samcrang:
Sounds reasonable

aws/modules/core/route53.tf
```diff
@@ -18,6 +18,14 @@ resource "aws_route53_record" "zone_delegation" {
   records = [ "${aws_route53_zone.core.name_servers}" ]
 }
 
+resource "aws_route53_record" "soa" {
+  zone_id = "${aws_route53_zone.core.zone_id}"
+  name = "${var.vpc_name}.${var.dns_domain}"
+  type = "SOA"
+  ttl = "900"
+  records = [ "${aws_route53_zone.core.name_servers.0}. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 ${var.soa_negative_cache_ttl}" ]
```
philandstuff: while we're here, do we want to change the email address to something we own? This might make it easier for TLS cert providers to verify certs, for example

samcrang: we chatted about this IRL and decided it's probably okay as it is for the time being.

PR #274, opened 2017-02-13
Serve a potentially different `robots.txt` in Beta

samcrang:
Currently, the app responds to `/robots.txt` with a restrictive
`robots.txt`. We want to have crawlers visit us in Beta only and only on
`*.register.gov.uk`. The CDN seems like the most sensible place to do
this.

### DIFF
--- a/aws/modules/register/cdn.tf
+++ b/aws/modules/register/cdn.tf
@@ -32,6 +32,28 @@ resource "aws_cloudfront_distribution" "distribution" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+  cache_behavior {
+    allowed_methods = ["HEAD", "GET"]
+    cached_methods = ["HEAD", "GET"]
+
+    path_pattern = "robots.txt"
+
+    default_ttl = 86400
+    max_ttl = 31536000
+    min_ttl = 0
+
+    forwarded_values {
+      cookies {
+        forward = "none"
+      }
+      query_string = true
+      headers = ["Accept", "Host", "Origin"]
+    }
+
+    target_origin_id = "${var.environment}-${var.name}-robots-txt"
+    viewer_protocol_policy = "redirect-to-https"
+  }
+
   origin {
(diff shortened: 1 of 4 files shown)


robmckinnon:
Which pages on `register.gov.uk` will be allowed for crawlers?

samcrang:
@robmckinnon currently it's no different to what's served on `openregister.org`. However, we're probabaly going to serve something like this (but it is still TBD):

```
User-agent: *
# Don't allow crawling too deeply into records or entries pages
Disallow: /records*?
Disallow: /entries*?

# Disallow for the moment, revisit later
Disallow: /proof
```

robmckinnon:
@samcrang Makes sense, as long as we include disallow rules for all possible record/item/entry pages. In addition to the above, think you'd need to include:

```
Disallow: /record*?
Disallow: /entry*?
Disallow: /item*?
```

samcrang:
@robmckinnon is there any particular reason why you feel we shouldn't allow these? I think the argument against `/records*?` and `/entries*?` is that so crawlers don't go wild on large registers (but I'm not sure I agree with doing this). I think there's value in having crawlers hit the things you've suggested?

robmckinnon:
@samcrang Currently GDS steers service teams to Disallow all on `*.service.gov.uk` pages. The motivation is for search engine results to point users to the service landing page on `www.gov.uk` to obtain context and caveats for the service, rather than a page within the service flow. e.g. service landing page: https://www.gov.uk/accelerated-possession-eviction

Are registers similar in that we may want people to hit a landing page per register in search results to get the context, but we wouldn't want them to hit a record page? e.g. https://school-eng.alpha.openregister.org/ versus a record page for "Harpenden Free School" https://school-eng.alpha.openregister.org/record/143648 - Particularly if there are other `gov.uk` pages that users would find more useful, e.g. compare schools page for "Harpenden Free School" https://www.compare-school-performance.service.gov.uk/school/138561

PR #279, opened 2017-02-14
migrate some of our instances to ubuntu

philandstuff:
This migrates some of our instances to ubuntu.  Specifically, the
even-numbered instances, when numbering from 0 (so where we have 3
instances, the first and last instance will be migrated to ubuntu).

This only migrates some of the instances so that we don't cause
downtime through this migration.  Once this has been deployed
everywhere, we can go back and migrate the rest.

Our user data (in users.yaml) needed very little changes to support
this.  All I did was:

 - rename some packages which are named differently on ubuntu (aws-cli
   to awscli, docker to docker.io)
 - remove yum-cron and its config (we have a card to reinstate
   ubuntu's equivalent)

I've preserved the original user data as
modules/register_group/old-users.yaml, to support keeping the old
Amazon Linux instances around.

Once this is merged, the remaining tasks are:

 - deploy this everywhere
 - migrate the bastion host
 - migrate the remaining register_group instances
 - clean up the interim code (ie remove old-users.yaml and the
   `element(list(...), count.index)` hacks in this commit)

### DIFF
--- a/aws/modules/register_group/old-users.yaml
+++ b/aws/modules/register_group/old-users.yaml
@@ -40,9 +40,10 @@ users:
     sudo: ALL=(ALL) NOPASSWD:ALL
 
 packages:
-  - awscli
-  - docker.io
+  - aws-cli
+  - docker
   - ruby
+  - yum-cron
 
 package_upgrade: true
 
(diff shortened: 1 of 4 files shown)


aws/modules/register_group/variables.tf
```diff
@@ -13,7 +13,7 @@ variable "vpc_id" {
 }
 
 variable "instance_ami" {
-  default = "ami-c51e3eb6"
+  default = "ami-98ecb7fe"
```
samcrang: How would you feel about using this?: https://www.terraform.io/docs/providers/aws/d/ami.html

aws/modules/register_group/old-users.yaml
```diff
@@ -0,0 +1,144 @@
+#cloud-config
+
+users:
+  - default
+
+  - name: chrisduraj
```
samcrang: Looks like `chrisduraj` could be removed from `users.yaml`?

aws/registers/templates/users.yaml
```diff
@@ -56,89 +55,7 @@ write_files:
     aws s3 cp s3://aws-codedeploy-eu-west-1/latest/install /tmp/install --region eu-west-1
     chmod +x /tmp/install
     /tmp/install auto
-- path: /etc/yum/yum-cron.conf
```
samcrang: Are we planning on doing the `unattended-upgrades` stuff in a different PR?

aws/modules/register_group/old-users.yaml
philandstuff: 2dbf9be

aws/registers/templates/users.yaml
philandstuff: done in @2011103

aws/modules/register_group/variables.tf
philandstuff: that looks interesting, maybe a separate pr though?

samcrang: 👍

PR #284, opened 2017-02-15
Specify instance image via aws_ami resource

philandstuff:
This commit specifies the instance image via the aws_ami resource.
This has the following advantages:

  - it makes it clearer which ami we're using, and what specific
    properties we want (in this case, ubuntu xenial on hvm with ssd
    storage)
  - it will automatically use newer AMIs when they become available

This has the following disadvantage:

  - it will automatically use newer AMIs when they become available

ie, when a new ubuntu AMI gets baked, this code will automatically
rebuild the world to use it.

Putting this out as a PR for discussion; I'm undecided about it.

### DIFF
--- a/aws/registers/main.tf
+++ b/aws/registers/main.tf
@@ -9,6 +9,15 @@ provider "statuscake" {
   apikey = "${var.statuscake_apikey}"
 }
 
+data "aws_ami" "ubuntu-xenial-ebs-ssd" {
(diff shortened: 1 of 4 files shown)


philandstuff:
Worth mentioning that this PR, at the moment, results in a plan which does no changes. Which is as expected 😄

PR #285, opened 2017-02-15
fix internal DNS

philandstuff:
This fixes our internal DNS naming.  In particular:

 - this sets all instance hostnames to match their names in the
   private route 53 hosted zone (for example,
   multi-1.test.openregister)
 - this sets DHCP options sets so that the DNS search domain makes
   sense for the environment (for example, test.openregister)
 - this adds a DNS record for the bastion (for example,
   bastion.test.openregister)

Before this change, each machine has a hostname that isn't
resolvable.  The hostname is set to something like ip-10-11-12-13, but
neither /etc/hosts nor the nameservers in /etc/resolv.conf can resolve
this into an IP address.  A number of tools get confused by
this (including the default sudo on ubuntu).

This also provides other benefits: for example, a more meaningful
hostname is more helpful in logs and metrics.


samcrang:
👎

samcrang:
Context: we had a chat about this and decided it's not worth the effort for solving the problem we see. The problem we see is `sudo` warning that it can't resolve the hostname which really isn't a big deal (maybe).

philandstuff:
The other thing we discussed is that this proposed solution wouldn't work if we moved to autoscaling 💩

PR #286, opened 2017-02-16
Scripts to load register and field registers

robmckinnon:
- Add script to load register register entries.
- Add script to load field register entries.
- Add usage examples of new loading scripts to readme.


samcrang:
👍

PR #289, opened 2017-02-16
Handle alternate data dir path in load script

robmckinnon:
- Change script to handle data paths that do not include phase name.
- Print register url at end of load script run.


samcrang:
👍

PR #290, opened 2017-02-16
make basic register_group instances t2.micro

philandstuff:
These don't need to be t2.medium.


philandstuff:
This has already been applied to test. #sns

karlbaker02:
Definitely.

PR #292, opened 2017-02-16
Ship cloudtrail to central bucket

philandstuff:


philandstuff:
ugh, this breaks because the multiple module instances don't play nice:

```
aws_iam_role.cloud_watch_logs_role: Error creating IAM Role CloudWatch-for-CloudTrail: EntityAlreadyExists: Role with name CloudWatch-for-CloudTrail already exists.
```
😞

PR #294, opened 2017-02-17
Add Telegraf configuration

samcrang:
Generating an stroing Telegraf configuration in S3 so we can sidecar a Telegraf container like we do with fluentd.


aws/modules/configuration/telegraf.tf
```diff
@@ -0,0 +1,18 @@
+resource "aws_s3_bucket_object" "telegraf_conf" {
+  bucket = "openregister.${var.vpc_name}.config"
+  key = "telegraf.conf"
+  content = "${data.template_file.telegraf.rendered}"
+  etag = "${md5(data.template_file.telegraf.rendered)}"
```
philandstuff: not [sha256](https://github.com/openregister/food-ratings-demo/pull/14)?

samcrang: We chatted about this IRL:

 > The entity tag is a hash of the object. The ETag reflects changes only to the contents of an object, not its metadata. The ETag may or may not be an MD5 digest of the object data. Whether or not it is depends on how the object was created and how it is encrypted as described below:
> -    Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-S3 or plaintext, have ETags that are an MD5 digest of their object data.
> -    Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-C or SSE-KMS, have ETags that are not an MD5 digest of their object data.
> -    If an object is created by either the Multipart Upload or Part Copy operation, the ETag is not an MD5 digest, regardless of the method of encryption.

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html

PR #299, opened 2017-02-23
Add custodian for `registration-district` in Discovery

samcrang:


philandstuff:
is it significant that no other discovery registers have a custodian name?

samcrang:
I did wonder this, I don't know. If the person in question is to be a custodian, I guess not? I'll double check though.

PR #305, opened 2017-02-27
Log git username in data load script slack msg

robmckinnon:
- To increase visibility of who is changing data.


robmckinnon:
Tested the script change a couple of times. Working well, so I'm going to merge.

PR #306, opened 2017-02-27
Automate some basic InfluxDB configuration

samcrang:
There's not a great deal of support for InfluxDB with Terraform but
something is probably better than nothing.

Whilst working on this I had looked at trying to unify a bunch of separately deployed things (for example: Sumo Logic lambdas, cloudtrail, TLS certs) and I ended up going down a bit of a rabbit hole where I ended up feeling that setting up `credstash` might make our lives easier. This is particularly relevant to the Sumo Logic Lambdas that require injecting secrets and zipping.

I think we should look at this soon, but not right now.


aws/influxcloud/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	plan-destroy
+
+export expected_terraform_version=v0.8
+export actual_terraform_version=$(shell terraform version | head -n 1 | cut -f2 -d" " | cut -d"." -f1,2)
+
+ifneq ($(expected_terraform_version), $(actual_terraform_version))
+  $(error Expected terraform version $(expected_terraform_version).x, but saw $(actual_terraform_version).x)
+endif
+
+export TF_VAR_cluster_address=$(or $(shell PASSWORD_STORE_DIR=~/.registers-pass pass services/influxcloud/admin | tail -n 1), $(error "Couldn't get InfluxCloud cluster address"))
```
philandstuff: `services/influxcloud/admin` looks like a copypaste error?

samcrang: I think this is right—it's getting the `tail -n1` from the file (which happens to the be the cluster address). The other usage of that credential gets `head -n1`.

samcrang:
I've found a better way to do this.

PR #311, opened 2017-02-28
Make `make plan` approximately twice as fast

samcrang:
It turns out that we've been assigning variables in `make` lazily. This
means that all variables are evaluated separately for each target.

For example, a `make plan` uses 4 targets. This means that `pass` is
called 4 times to get the same password. This is unacceptable for us
because decryption operations on `pass` are slow. We can measure this
with this with the following:

```
{ time make plan -e vpc=test } 2>> results.txt
```

Before (mean wall time: ~85s):

```
make plan -e vpc=test  19.60s user 6.60s system 30% cpu 1:26.22 total
make plan -e vpc=test  20.19s user 6.96s system 23% cpu 1:55.05 total
make plan -e vpc=test  19.82s user 6.91s system 29% cpu 1:29.93 total
make plan -e vpc=test  19.45s user 6.47s system 35% cpu 1:13.01 total
make plan -e vpc=test  19.38s user 6.44s system 40% cpu 1:03.06 total
```

After (mean wall time: ~41s):

```
make plan -e vpc=test  15.15s user 4.15s system 42% cpu 44.942 total
make plan -e vpc=test  14.94s user 4.10s system 41% cpu 45.683 total
make plan -e vpc=test  15.01s user 4.13s system 49% cpu 38.784 total
make plan -e vpc=test  15.10s user 4.13s system 48% cpu 39.696 total
make plan -e vpc=test  15.06s user 4.29s system 47% cpu 40.740 total
```

### DIFF
--- a/aws/cloudtrail/Makefile
+++ b/aws/cloudtrail/Makefile
@@ -5,17 +5,17 @@
 	plan \
 	plan-destroy
 
-export expected_terraform_version=v0.8
-export actual_terraform_version=$(shell terraform version | head -n 1 | cut -f2 -d" " | cut -d"." -f1,2)
+export expected_terraform_version:=v0.8
+export actual_terraform_version:=$(shell terraform version | head -n 1 | cut -f2 -d" " | cut -d"." -f1,2)
 
 ifneq ($(expected_terraform_version), $(actual_terraform_version))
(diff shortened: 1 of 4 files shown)


samcrang:
Also this might be particularly slow for me because I have a large key on a Yubikey.

PR #315, opened 2017-03-03
Log `Host` and `X-Forwarded-Host`

samcrang:
Without logging these it is not possible to see which Register was
requested.

For future reference the log conversions are documented here:

https://logback.qos.ch/manual/layouts.html#logback-access

This format lets us use `keyvalue` in Sumo Logic for easy parsing:

https://help.sumologic.com/Search/Search_Query_Language/01_Parse_Operators/04_Parse_Keyvalue_Formatted_Logs


samcrang:
I think we should reject this and go with using [alphagov/dropwizard-logstash](https://github.com/alphagov/dropwizard-logstash/) instead.

PR #323, opened 2017-03-20
Fix `rm` flags

samcrang:
Looks like I managed to mistype some flags :weary:


endamccormack:
I think `aws/tls-certs/Makefile` will still contain `rm -r .terraform/terraform.tfstate` after this change.

samcrang:
Hmm you're not wrong.

samcrang:
Fixed it. Weirdly `ag "rm -r"` didn't catch this for me but `git grep "rm -r"` (which I don't use) does 🤔.

PR #325, opened 2017-03-20
Enable ELB connection draining

samcrang:
I can't see any good reason why we don't enable this. Whilst we don't
add/remove instances from the ELB automatically, this will make world
rebuilds, machine reboots etc. have less of an impact.


samcrang:
Looking into it a bit more I'm not sure this will make a difference for the scenarios I descsribed. It will make a difference for situations where we deregistering instances for some reason, perhaps to debug weirdness on a box, something I've been doing recently.

Either way, it shouldn't have a negative impact.

PR #326, opened 2017-03-21
Stop performing reverse DNS lookups of RDS clients

samcrang:
It turns out this was causing slowness for our RDS clients. This was
most pronounced when deploying but almost certainly affected most DB
connections in some way.

In a normal situation you would expect these reverse lookups to be fast.
However, because we accidentally chose "private" IP ranges within
`172.0.0.0/8` but outside of `172.16.0.0/12` we ended up assigning
privately IPs in the public address space. This means when Postgres
performs a reverse DNS lookup we rely on the owners of those IP address
to have configurations that respond with timely responses.

Frequently this is unproblematic. I imagine in some instances (although
I've not observed this personally) we'll see correct, affirmative
responses. In other case we've seen `NXDOMAIN` responses which should be
cached based off the `SOA` record. The slow ones I've personally
observed have responded with `SERVFAIL`; these are slow and are probably
not cached by recursors (§7.1 of RFC2308) so the failure has a bigger
impact.

Ultimately, we should look to fix the issue of misallocated private IP
ranges, but in the short-term this workaround should stop us from seeing
problems.

### DIFF
--- a/aws/modules/database/parameter_group.tf
+++ b/aws/modules/database/parameter_group.tf
@@ -1,0 +1,10 @@
+resource "aws_db_parameter_group" "db" {
+  name   = "${var.id}"
+  family = "postgres9.5"
+
+  parameter {
+    name = "log_hostname"
+    value = false
+    apply_method = "immediate"
+  }
+}
--- a/aws/modules/database/main.tf
+++ b/aws/modules/database/main.tf
@@ -16,7 +16,7 @@ resource "aws_db_instance" "db" {
   engine = "${var.engine}"
 
   db_subnet_group_name = "${aws_db_subnet_group.subnet_group.id}"
-  parameter_group_name = "${var.parameter_group_name}"
+  parameter_group_name = "${aws_db_parameter_group.db.id}"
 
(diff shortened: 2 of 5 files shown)


aws/modules/database/parameter_group.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_db_parameter_group" "db" {
+  name   = "${var.id}"
+  family = "postgres9.5"
+
+  parameter {
+    name = "log_hostname"
+    value = "0"
```
endamccormack: Can we do off instead of 0 to make it more clear?

https://gpdb.docs.pivotal.io/4370/guc_config-log_hostname.html

samcrang: Possibly. It's represented in the AWS console as `0` and `1` but I'll see if deals with other values.

samcrang:
@endamccormack how's this? Setting `"on"` or `"off"` apply successfully but subsequent plans continue to think there's changes to apply. I've gone for `false` because these are represented inside Terraform as the strings `"0"` and `"1"` and express the intention more clearly.

PR #329, opened 2017-04-07
Add deployment pipeline for `openregister-java`

samcrang:
We still need to change `.travis.yml` in `openregister-java` so it puts the artifact in the correct place for CodePipeline.

### DIFF
--- a/aws/pipeline/openregister-java_pipeline.tf
+++ b/aws/pipeline/openregister-java_pipeline.tf
@@ -1,0 +1,169 @@
+resource "aws_iam_role" "codepipeline" {
+  name = "codepipeline"
+
+  assume_role_policy = <<EOF
+{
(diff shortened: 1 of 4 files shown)


aws/pipeline/openregister-java_pipeline.tf
```diff
@@ -0,0 +1,169 @@
...
+    action {
+      name = "TravisBuildOutput"
+      category = "Source"
+      owner = "AWS"
+      provider = "S3"
+      version = "1"
+      output_artifacts = ["openregister_package"]
+
+      configuration {
+        S3Bucket = "openregister.app.config"
```
johnollier: should be openregister.app.artifacts

samcrang: Good spot 🔍

PR #332, opened 2017-04-10
Allow instances to add/removes themselves from ELBs

samcrang:
This will allow us to push the responsibility of draining connections onto the ELB and therefore support zero-downtime deploys.


karlbaker02:
Looks good

PR #333, opened 2017-04-10
Explicitly specify the ELB connection draining timeout

samcrang:
This value will become more important as we rely on ELBs to do
connection draining during deploys. We should be explicit about what the
value is rather than using the AWS default.


karlbaker02:
👍

PR #341, opened 2017-04-26
Reduce number of metrics sent to influxdb

samcrang:


karlbaker02:
Sam talked the changes through with me, all sounded good.

PR #350, opened 2017-05-04
New alpha registers

michaelabenyohai:


ansible/group_vars/tag_Environment_alpha
```diff
@@ -55,26 +58,17 @@ register_groups:
...
-    - religious-character
+    - principal-local-authority
+    - prison-estate
     - registration-district
-    - school-admissions-policy
-    - school-authority-eng
     - school-eng
-    - school-gender
-    - school-phase
-    - school-tag
```
samcrang: I take it all these school related things are related to moving everything into `school-eng`?

michaelabenyohai: They have remove those as fields completely from school-eng and also removing the registers. See https://github.com/openregister/registry-data/pull/80.

Also, I have not removed the school-* registers from `registers.tf` yet as it is likely that they might end up in alpha at some point.

samcrang: 👍

PR #353, opened 2017-05-08
Send Statuscake availability to Performance Platform

samcrang:


samcrang:
This currently fails if the mean uptime is not an integer. For example:

```{'messages': ["99.996 is not of type u'integer'"]}```

samcrang:
Mergin' this beceause me and @johnollier paired on it.

PR #355, opened 2017-05-11
Send logs as JSON

samcrang:
https://github.com/openregister/openregister-java/pull/276 needs to get merged first.

### DIFF
--- a/ansible/roles/openregister_config/templates/openregister-config.yaml.j2
+++ b/ansible/roles/openregister_config/templates/openregister-config.yaml.j2
@@ -58,6 +58,10 @@ downloadConfigs: true
 
 server:
   registerDefaultExceptionMappers: false
+  requestLog:
+    type: classic
+    appenders:
+      - type: logstash-console
 
 credentials:
   user: {{ app_user }}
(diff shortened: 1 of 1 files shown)


ansible/roles/openregister_config/templates/openregister-config.yaml.j2
```diff
@@ -58,6 +58,10 @@ downloadConfigs: true
 
 server:
   registerDefaultExceptionMappers: false
+  requestLog:
+    type: classic
```
endamccormack: I cant see anything about this type attribute in the documentation. Whats it for?

samcrang: hmmm, I _think_ it's related to the format that logback uses but I'll double check and see what it's actually doing.

samcrang: I don't entirely understand why it's needed but without it it fails to log requests /shrug

samcrang: @endamccormack https://github.com/dropwizard/dropwizard/blob/master/docs/source/manual/configuration.rst#classic-request-log

PR #383, opened 2017-07-06
Send availability from Pingdom to Performance Platform

samcrang:


aws/lambda/python/pingdom-to-performance-platform/README.md
```diff
@@ -0,0 +1,15 @@
...
+```bash
+$ python3 -m unittest
+```
+
+## Running Locally
+
+You can run it locally to generate some output (does not publish to Performance Platform):
+
+```bash
+PINGDOM_KEY=$(registers-pass show services/pingdom/api | head -n1) PINGDOM_USERNAME=$(registers-pass show services/pingdom | tail -n 1) PINGDOM_PASSWORD=$(registers-pass show services/pingdom| head -n1) performance_platform.py 2>/dev/null
```
johnollier: One tiny thing - should be ./performance_platform.py ?

samcrang: yes it should be. 👍

PR #390, opened 2017-07-18
Script to create RSF and load in ORJ

johnollier:
This consists firstly of a python script that generates RSF from a) register and fields data which is read either from an existing register e.g. field.alpha.openregister.org or from the file system b) user data in a TSV file, YAML file or set of YAML files in a single directory.

The python script is called by various bash scripts which either:
a) deletes an existing register data and loads a complete register including metadata from a TSV file
b) updates a register without deleting existing data or adding metadata from a TSV file
c) deletes an existing metadata register and reloads the complete register including metadata from a directory of YAML files
d) updates an existing metadata register e.g. adding a new register to the Register Register from a YAML file.


johnollier:
@michaelabenyohai has reviewed up to bbb772d and the next commit contains the implementation of her suggestions. There were a couple of things to fix after using the scripts to deploy statistical-geography.discovery. We have discussed in standup and everyone is happy to merge this.

PR #393, opened 2017-07-26
Send Docker logs to Logit.io

samcrang:


aws/registers/templates/fluentd.conf
```diff
@@ -52,4 +52,13 @@
     endpoint https://endpoint1.collection.eu.sumologic.com/receiver/v1/http/${sumologic_key}
     log_format json
   </store>
+
+  <store>
+    type logit
+    stack_id ${logit_stack_id}
+    port 16081
```
philandstuff: is it worth making the port number configurable too? (these do vary with stack id)

samcrang: Ahhh, I didn't realise this. I clearly made a bunch of assumptions about this being the `TCP-SSL` port. I'll extract it to some configuration.

PR #396, opened 2017-08-03
Stop "anonymising" IPv4 logs

samcrang:
This is a very crude and not particularly reliable or valuable way of
"anonymising" IPv4 addresses. We've decided to remove it because we do
not think it's required. It also didn't work anyway because the original
message field remained unchanged.

This also resolves the problem of logs not getting sent to Sumologic/Logit.


samcrang:
We need to rethink this because the anonymisation is required.

PR #400, opened 2017-08-08
Create new alpha registers

michaelabenyohai:
- government-domain
- social-housing-provider-designation
- social-housing-provider-legal-entity


karlbaker02:
Do we want to remove `government-domain` from discovery in this commit too, or shall we allow them to co-exist for a while?

michaelabenyohai:
@karlbaker02 I was planning on doing that in a separate commit later today or tomorrow. Just in case anything goes wrong.

karlbaker02:
Sounds good @michaelabenyohai  :)

PR #402, opened 2017-08-09
Fix canonicalization

karlbaker02:
This PR fixes canonicalization with the `rsfcreator.py` script so that empty key-value pairs are not included in metadata items and to stop escaping special characters as unicode.

### DIFF
--- a/scripts/rsfcreator.py
+++ b/scripts/rsfcreator.py
@@ -86,8 +86,10 @@ def generate_rsf(args):
             custodian_item_line, custodian_entry_line = rsf_for_line({'custodian': args.custodian}, 'custodian', 'system', key='custodian')
             print(custodian_item_line + '\n' + custodian_entry_line)
         for field in fields_by_name.values():
-            field_item_line, field_entry_line = rsf_for_line(field, 'field', 'system', key_prefix='field:')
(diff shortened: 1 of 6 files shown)


scripts/rsfcreator.py
```diff
@@ -55,7 +55,7 @@ def rsf_for_line(line_dict, key_field, entry_type, key_prefix='', key=None):
     if key is None:
         key = line_dict[key_field]
     timestamp = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
-    item_str = json.dumps(line_dict, separators=(',', ':'), sort_keys=True)
+    item_str = json.dumps(line_dict, ensure_ascii=False, separators=(',', ':'), sort_keys=True)
```
gidsg: Are these fixes testable, are they worth testing?

michaelabenyohai: Good point. We do have a test class for `rsfcreator.py`. Can we add some tests @karlbaker02 ?

karlbaker02: Yes, I'll do this, thanks.

karlbaker02: Test has now been added.

PR #412, opened 2017-08-21
Use correct logstash appender for PaaS/AWS

michaelabenyohai:
The access-logstash-console appender does some useful parseing
of fields. However, this is incompatible with fluentd, which
consumes logs when running the application on AWS. We therefore
need to make sure we use the correct appender for the platform
on which we are running the application.

### DIFF
--- a/ansible/roles/openregister_config/templates/openregister-config.yaml.j2
+++ b/ansible/roles/openregister_config/templates/openregister-config.yaml.j2
@@ -66,8 +66,14 @@ server:
       port: {% if paas %}${PORT}{% else %}8080{% endif %}
   registerDefaultExceptionMappers: false
   requestLog:
+    {% if paas %}
     appenders:
       - type: access-logstash-console
+    {% else %}
+    type: classic
+    appenders:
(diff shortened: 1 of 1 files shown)


ansible/roles/openregister_config/templates/openregister-config.yaml.j2
```diff
@@ -66,8 +66,9 @@ server:
       port: {% if paas %}${PORT}{% else %}8080{% endif %}
   registerDefaultExceptionMappers: false
   requestLog:
+    type: classic
     appenders:
-      - type: access-logstash-console
```
samcrang: This change should only be applied when non running on PaaS. There's a magical `paas` bool you can use to make the generated configs different.

michaelabenyohai: Ah yes. I had forgotten that we were still using config from S3 for PaaS! Will update.

michaelabenyohai:
@samcrang I've addressed this now.

PR #416, opened 2017-08-25
Swap DNS to PaaS-backed infrastructure

samcrang:
This will move away from our self-managed AWS infrastructure to PaaS.

Things to double-check before deploying: Terraform shows that this will be resource modifications. We need to double check that this isn't modelled as a delete/create so that we won't end up with any records that are unresolvable for a period of time.


samcrang:
I had a quick look at this and it appears that the DNS changes happen as either upserts or transactional changes to Route53 which should mean that we see zero downtime when swapping over.

PR #417, opened 2017-08-29
Move address to multi tenancy group

samcrang:
Aside from the fact it's unnecessary for the address register to have its own instances this also simplifies the migration of data to PaaS.

I'm happy to migrate the data from the address group to the multi group.


karlbaker02:
I think that given the small size of `address` in `discovery`, this is fine. If we start to grow address again to the millions of records it was previously, we may need to revisit this decision.

ansible/group_vars/tag_Environment_discovery
```diff
@@ -15,8 +15,6 @@ register_settings:
     enable_register_data_delete: false
     fields_yaml_location: "s3://openregister.discovery.config/fields.yaml"
     registers_yaml_location: "s3://openregister.discovery.config/registers.yaml"
-  address:
```
karlbaker02: We should still keep this configuration option.

samcrang: @karlbaker02 I thought we only had this enabled because the register was large? Is there another reason?

karlbaker02: I thought there was some legal issue around address data? I may be mistaken, in which case we could remove it.

samcrang: @karlbaker02 that's possibly true, however, the register only contains 833 records and you could pull all that in one request from the `/records` endpoint within the `page-size` limit anyway.

michaelabenyohai: This was only there because the data was large and a download made the app fall over. This won't happen for the small address register in discovery.

PR #423, opened 2017-09-07
Remove history page link from register homepages

fjavierm:


gidsg:
Is it safe to remove this usage also? https://github.com/openregister/deployment/blob/e48989bf81062ce8365a83247707bd388675554f/ansible/roles/openregister_config/templates/openregister-config.yaml.j2#L108

fjavierm:
The issue is just talking about to remove them from the config generation. It says explicitly not to remove the code that make them configurable.

PR #431, opened 2017-09-27
remove unrequired registers from alpha

gidsg:


aws/registers/registers.tf
```diff
@@ -232,24 +232,6 @@ module "green-deal-certification-body_register" {
   pingdom_contact_ids = "${var.pingdom_contact_ids}"
 }
 
-module "government-domain_register" {
```
michaelabenyohai: You can't remove these modules because some of these registers exist in other environments (beta) and these modules are still required when applying to those environments.

The only one you can remove is `school-type_register` module because this no longer exists in any environment (it was renamed to school-type-eng).

PR #436, opened 2017-10-12
remove bastion host

gidsg:
remove AWS bastion host as this is no longer needed.


gidsg:
Note, I have run this as `make plan` only. I was thinking to apply it once merged.

PR #454, opened 2017-11-17
specify python version

### LINKED ISSUE #446: specify python2 is needed for virtualenv

gidsg:
We need to specify usage of Python 2 if user has multiple python versions installed. fixes #446


fjavierm:
I agree with Arnau. We should specify the version dependencies.

PR #458, opened 2017-11-27
Add average response time and location as metrics to performance platform data

karlbaker02:
This PR adds two new metrics - `responsetime` (from the `avgresponse` provided by Pingdom) and `location` - which will be sent alongside the existing payload to the Performance Platform. Splitting out `location` from the name of the check will make filtering easier, whilst `responsetime` is added to as additional data to display on the report.

### DIFF
--- a/aws/lambda/python/pingdom-to-performance-platform/performance_platform.py
+++ b/aws/lambda/python/pingdom-to-performance-platform/performance_platform.py
@@ -98,16 +98,18 @@ def build_payload(pingdom, tag):
         timestamp = start_of_previous_hour.strftime("%Y-%m-%dT%H:00:00+00:00")
         uptime = summary["summary"]["status"]["totalup"]
         downtime = summary["summary"]["status"]["totaldown"]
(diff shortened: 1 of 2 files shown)


karlbaker02:
One thought I have had is whether it is correct to upload the response time data in the same payload as the uptime data, or whether this data should go to a separate endpoint (and if it does go to a separate endpoint, is this something we can manage ourselves or will the Performance Platform folk need to create it for us)?

karlbaker02:
To answer my own comment, let's improve this later on and for now get the page live.

michaelabenyohai:
I assume this means that only the latest updates to Performance Platform will contain the new attributes? So we won't be able to show historical data?

karlbaker02:
@michaelabenyohai That's correct, unless we were to somehow replay all of our historical data back from Pingdom (assuming they keep this data) to Performance Platform.

michaelabenyohai:
@karlbaker02 I'm happy if @elliecraven is happy so let's go for it 👍

PR #462, opened 2017-12-04
update let's encrypt docs

gidsg:
Update docs based with latest info based on me trying to follow the README!


README.md
```diff
@@ -226,11 +226,18 @@ to delete the old certificate from IAM.
...
-Use the AWS API Gateway console to rotate the SSL certificate. 
-The AWS Console for API Gateway has changed recently and it is only possible to use certificates stored 
-in AWS Certificate Manager in the US East (N. Virginia) `us-east-1` for a Custom Domain Name. Using
-the Certificate Manager console (in the US East (N. Virginia) region) reimport the certificate with domain
-name `register.gov.uk`. 
+Use the AWS CLI to rotate the SSL certificate.
+
+Get the Certificate ARN by running. 
+`aws acm list-certificates --region us-east-1`
+and copy the ARN for `register.gov.uk`
```
michaelabenyohai: Should we mention that this needs to be the latest certificate? (There may be two of them if the delete in the previous step failed).

gidsg: I *think* you may be confusing the IAM ID (which will change) with the ACM ARN which should not change (you are just changing the certificate associated with it). I have added a note to check the upload date for IAM though.

michaelabenyohai: Ah ok, yes I was confusing it. Thanks!

PR #473, opened 2018-01-18
Remove old discovery registers

danmitchell-:
To match https://github.com/openregister/openregister-java/pull/418


michaelabenyohai:
Looks good apart from `place` being removed.

danmitchell-:
@michaelabenyohai I have put place back :)

PR #479, opened 2018-01-29
Change discovery registers to point at new discovery space

karlbaker02:
### Context
This PR updates CodePipeline to deploy discovery registers to the new `discovery` space, rather than to `prod`.


michaelabenyohai:
This looks good. The only problem is it will fail to deploy correctly because [buildspec.yml](https://github.com/openregister/openregister-java/blob/master/deploy/buildspec.yml) will deploy it to an app called `discovery-multi`/`discovery-basic`. We might need to rename the apps.

PR #481, opened 2018-01-30
Add readme for making disco register environments

nacnudus:
### Context
Non-devs are to create their own discovery register environments.  This PR gives the instructions.

Add a markdown document of instructions that GitHub will render nicely, but that won't replace the default README.md.

### Guidance to review
Check that all the necessary steps to create a new discovery register are clearly explained in the right order.

### DIFF
--- a/create-a-discovery-register-environment.md
+++ b/create-a-discovery-register-environment.md
@@ -2,22 +2,20 @@
 
 In these instructions, the new register to be created is called `meat-cuts`
 
+## Prerequisites
+
+- A GOV.UK PaaS account and access to the `openregister` organisation and `discovery` space as `SpaceDeveloper`.
+- Access to the AWS console and [awscli](https://aws.amazon.com/cli/) installed (you will need to create and use an API key).
+
 ## 1. AWS: edit the config file
 
 Tell everyone you're creating a new register, so that you don't both do it at
 the same time.
 
-Log into AWS https://openregister.signin.aws.amazon.com/console
+Download the latest config file from S3, which is used by the application to
+know all the metadata about the register, using `./scripts/download-s3-config.sh`.
 
-Go to the [S3 bucket](https://s3.console.aws.amazon.com/s3/buckets/openregister.discovery.config/?region=us-east-2&tab=overview)
-or browse to `Services > Storage > S3 > openregister.discovery.config`
-
-Download this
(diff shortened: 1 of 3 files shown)


create-a-discovery-register-environment.md
```diff
@@ -0,0 +1,83 @@
...
+Tell everyone you're creating a new register, so that you don't both do it at
+the same time.
+
+Log into AWS https://openregister.signin.aws.amazon.com/console
+
+Go to the [S3 bucket](https://s3.console.aws.amazon.com/s3/buckets/openregister.discovery.config/?region=us-east-2&tab=overview)
+or browse to `Services > Storage > S3 > openregister.discovery.config`
+
+Download this
+[file](https://s3.console.aws.amazon.com/s3/object/openregister.discovery.config/multi/openregister/paas-config.yaml?region=us-east-2&tab=overview)
```
gidsg: might be easier to do this with the AWS CLI instead? https://docs.aws.amazon.com/cli/latest/reference/s3/index.html we can then wrap these in shell scripts like `download_config.sh` or something?

create-a-discovery-register-environment.md
```diff
@@ -0,0 +1,83 @@
...
+change the step `registers-pass generate -n os/dev/test 16` to `registers-pass
+generate -n discovery/app/mint/meat-cuts 16`
+
+Once the registers-pass pull request has been merged, pull the registers-pass
+repository again and then use `registers-pass show discovery/app/mint/meat-cuts`
+to show the new password.  Put it into the `paas-config.yaml` file that you are
+still editing.
+
+```yaml
+  fire-authority:
```
michaelabenyohai: You also need to mention that the key of this section needs to be changed from `fire-authority` to `meat-cuts`

create-a-discovery-register-environment.md
```diff
@@ -0,0 +1,83 @@
...
+cf restart discovery-multi
+```
+
+When it restarts, it checks the config file, which is why you have to do that
+first.
+
+It will take everything down for a minute or so.
+
+If the name has already been created, then the `cf map-route` step will fail,
+but you should have seen it in `cf routes` first anyway.
```
michaelabenyohai: This bit isn't quite true. It will fail if the that route exists anywhere else in the whole of PaaS. i.e. it will fail if https://meat-cuts.cloudapps.digital has been claimed by any other user of PaaS.

We are hoping that this is unlikely to happen, but if it does speak to a developer and they can help by giving it a different domain via AWS.

PR #487, opened 2018-02-09
Allow creation of metadata RSF

karlbaker02:
### Context

This PR expands the capabilities of the `load-register-tsv.sh` script, adding the ability for the user to specify the type of data they wish to populate the generated RSF file. This can either by metadata, user data or both metadata and user data.

### Guidance for review

Tests need to be added


scripts/reload-metadata-yaml.sh
```diff
@@ -32,9 +32,9 @@ update_data_repo 'registry-data'
 echo "converting $YAML to RSF"
 if [ "$METADATA_SOURCE" = 'local' ]
 then
-  python3 $OPENREGISTER_BASE/deployment/scripts/rsfcreator.py $REGISTER $PHASE --yaml_dir $YAML --prepend_metadata --custodian "${CUSTODIAN}" --register_data_root $OPENREGISTER_BASE > $OPENREGISTER_BASE/tmp.rsf
+  python3 $OPENREGISTER_BASE/deployment/scripts/rsfcreator.py $REGISTER $PHASE --yaml_dir $YAML --prepend_metadata --include_user_data --custodian "${CUSTODIAN}" --register_data_root $OPENREGISTER_BASE > $OPENREGISTER_BASE/tmp.rsf
```
michaelabenyohai: Should we include the `[metadata|user|all]` option for this script too?

scripts/load-register-tsv.sh
```diff
@@ -9,7 +9,7 @@ source "$OPENREGISTER_BASE/deployment/scripts/includes/set-vars.sh"
 
 usage()
 {
-  echo "usage: ./load-register-tsv.sh [register] [phase] [tsvfile relative to root] [local|remote] [custodian] [optional data dir relative to root if not REGISTER-data]"
+  echo "usage: ./load-register-tsv.sh [register] [phase] [tsvfile relative to root] [metadata|user|all] [local|remote] [custodian] [optional data dir relative to root if not REGISTER-data]"
```
michaelabenyohai: I'm wondering whether it's worth calling this `system` instead of metadata, just to be consistent with the entry type?

karlbaker02: Sure.

scripts/reload-metadata-yaml.sh
karlbaker02: We could do. I didn't because the script at the moment is intended for reloading registers, which means that you'll always need the system entries first. That being said, the `load-register-tsv.sh` script also deletes all register data before loading.

My intention with this script is to make it simpler to generate the metadata for a register separately from the actual data. It may be that we need some more control when invoking these scripts to determine, say, whether or not to go ahead and delete data from a register before they become more useful.

michaelabenyohai: Yes it looks like `load-register-tsv.sh` and `reload-metadata-yaml.sh` both do the same thing but for different registers. They both ask (yes/no) whether you would like to delete data from the register first, so we do already have control over whether they delete the data. Or did you mean something else?

I think it would be good to make both scripts allow a user to specify `[metadata|user|all]` and maybe rename `reload-metadata-yaml.sh` to `load-metadata-yaml.sh` so that it's more obvious that they do similar things.

Also, the changes to `load-register-tsv.sh' now make the `update-register-tsv.sh` script redundant so can we remove it?

karlbaker02: `load-register-tsv.sh` and `reload-metadata-yaml.sh` will both delete data from the register before reloading it - neither of these scripts provides you an option of _would you like to delete the data?_. 

In terms of purpose, you're right that these two scripts are doing the same thing for different registers: `load-register-tsv.sh` loading data for regular registers, `reload-metadata-yaml.sh` loading data for the metadata registers.

If we were to make the deletion of data from the register an option to the user, then I can see the value in having `[metadata|user|all]` options to determine which data to load into the register, for both of these scripts.

Finally, if we made the change above, then we could safely remove `update-register-tsv.sh` as it's functionality would be covered by `load-register-tsv.sh`. As it stands, this is not the case because of the fact that deleting or not deleting register data is hard-coded into the scripts and not available as user input.

PR #498, opened 2018-03-13
add cloudwatch logs to Google Analytics Lambda

gidsg:
### Context
Part of the API Key spike

Add Lambda which sends CloudWatch Logs events to Google Analytics

### Guidance to review
Test event for this function can be invoked from the AWS Lambda Console.


danmitchell-:
Is this going into “production”? If so, should we want to remove the console.logs?

gidsg:
@danmitchell- As this is a server side Lambda `console.log` statements aid greatly with debugging, so I'd rather leave them in. Also there are none of the browser support concerns associated with client-side JS.

danmitchell-:
Can we run these through [XO](https://github.com/xojs/xo )

aws/lambda/node/log-api-key-to-cloudwatch/index.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const AWS = require('aws-sdk');
+const dataFormatRegex = /(csv|tsv|json|yaml|ttl)$/;
+
+exports.handler = (event, context, callback) => {
+   const request = event.Records[0].cf.request;
+   
+   // Let request continue immediately then carry out our logic
+   callback(null, request);
```
gidsg: Am not _sure_ this actually does anything? As per:
> When the callback is called, the Lambda function exits only after the Node.js event loop is empty (the Node.js event loop is not the same as the event that was passed as a parameter). 

https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html

karlbaker02: I can confirm that we definitely need to call the callback, but agree that it can probably be moved to the end of the function because the function definitely does not terminate at this point.

gidsg:
@danmitchell- I ran mine through `standardjs` as this seems to be the GDS approved linter as per: https://github.com/alphagov/styleguides/blob/master/js.md#linting ?

karlbaker02:
I ran mine through XO but happy to run it through `standardjs` if that's what we need to do.

danmitchell-:
XO is the new gds way so yours is fine @karlbaker02 👍

gidsg:
@karlbaker02 Sorry, turns out XO is the new recommended tool as per: https://gds-way.cloudapps.digital/manuals/programming-languages/nodejs/#source-formatting-and-linting

PR #499, opened 2018-03-14
removed `name` from the 'Loading a new field or register' section

trvllngmn:
i removed `name` from the 'Loading a new field or register' section,
from the argument list, the first argument, I removed 'name' from 'register/field name'.
In order to make it clear that only a distinction of 'register' or 'field' is needed.


karlbaker02:
Hi @trvllngmn, thanks for your input! The script does actually need the _name_ of the new field or register to be loaded, instead of this only being a distinction between whether a register _or_ field is to be loaded.

PR #501, opened 2018-03-29
Terraform api key

karlbaker02:
### Context

This PR updates our Terraform configuration to add the necessary resources for logging API key usage. This consists of:

- Lambda@Edge function
- Lambda function
- New role and policy for the Lambda@Edge function
- Associating CloudFront distributions with the Lambda@Edge function
- New log groups in each region for the Lambda@Edge function

The Lambda@Edge function can only exist in `us-east-1`. The other Lambda function and log groups are created on a per-region basis.


gidsg:
Note, there are many whitespace changes in this PR caused by a terraform linter so it is better to review excluding whitespace.

PR #503, opened 2018-05-02
Add redirects

michaelabenyohai:
### Context
Create redirects to the new service domain.

Create 3 redirect apps as per https://docs.cloud.service.gov.uk/#redirecting-all-traffic.

### Guidance to review
Review the README makes sense and that you can run it.

### DIFF
--- a/paas-redirect/app/index.html
+++ b/paas-redirect/app/index.html
@@ -1,8 +1,0 @@
-<html>
-  <head>
-    <title>Static Site</title>
-  </head>
-  <body>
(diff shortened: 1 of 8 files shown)


paas-redirect/app/index.html
```diff
@@ -0,0 +1,8 @@
+<html>
```
gidsg: this file is kinda strange. Does it still work if we make this an empty file?

michaelabenyohai: Yes it does - it's empty now 👍

PR #504, opened 2018-05-02
randomise CID

gidsg:
### Context
Suggested by Dan Gilbert

Randomise CID to avoid GA single user throttle

### Guidance to review
Data should appear in GA


gidsg:
lots of whitespace so better to preview changes with `w=1`

PR #506, opened 2018-05-14
Remove cloudfront-s3-logs-to-sumologic lambda

michaelabenyohai:
### Context
We no longer send these to Sumologic and instead send them to Logit.

Remove cloudfront-s3-logs-to-sumologic lambda.

### Guidance to review
`make plan` should plan to destroy the `aws_lambda_function.cloudfront_logs_to_sumologic` resource.


gidsg:
can we also safely remove the associated terraform variable now? e.g. https://github.com/openregister/deployment/blob/64ca8d09211a4631821e5fc8b406a7bf4c4bf5ce/aws/modules/configuration/variables.tf#L3
https://github.com/openregister/deployment/blob/64ca8d09211a4631821e5fc8b406a7bf4c4bf5ce/aws/modules/configuration/fluentd.tf#L12

michaelabenyohai:
@gidsg good spot. Removed those now, see updated comment.

PR #508, opened 2018-06-05
add option to add system entry updating field text

gidsg:
### Context
Previously we did not have a programatic way to generate field text updates. i.e. those enabled by https://github.com/openregister/openregister-java/pull/332

Add option to `update-metadata-yaml.sh` to write field updates to non-basic register
Note: I wrote the script in Ruby because I found it easiest but if people feel strongly against this they are welcome to help me port it to another language 😉 

### Guidance to review
A usage example is:
`./update-metadata-yaml.sh field test registry-data/data/beta/field/start-date.yaml local country`
which would update `start-date` in the `country` register.

### DIFF
--- a/scripts/system_field_text_update.rb
+++ b/scripts/system_field_text_update.rb
@@ -1,0 +1,2 @@
+rsf = IO.read(ARGV[0])
+File.write(ARGV[0], rsf.gsub!(/\t(user)\t((\w|-)+)/m, "\tsystem\tfield:" + '\2'))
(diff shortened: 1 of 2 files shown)


scripts/system_field_text_update.rb
```diff
@@ -0,0 +1,2 @@
+rsf = IO.read(ARGV[0])
+File.write(ARGV[0], rsf.gsub!(/\t(user)\t((\w|-)+)/m, "\tsystem\tfield:" + '\2'))
```
arnau: If I'm not mistaken, you can remove the parenthesis wrapping `user` and avoid the third group capture by using `[]`:

```
File.write(ARGV[0], rsf.gsub!(/\tuser\t([\w-]+)/m, "\tsystem\tfield:" + '\1'))
```

PR #512, opened 2018-06-21
Add name-a-register script

nacnudus:
### Context

Registers can now have titles independently of their unique ID.  This script is
for setting that title.

It might make more sense to incorporate the register title into the yaml
definitions of register metadata, but it was quicker to make this script.

Add a script `/scripts/name-a-register.sh` for naming a register as follows:

```sh
./name-a-register.sh alpha country "Countries recognised by the UK"
```

### Guidance to review

1. Check out the branch
2. Test the script with, e.g. the `fire-authority` register in the `alpha` environment
3. Download the RSF of the register and check that the `register-name` entry exists

### DIFF
--- a/scripts/name-a-register.sh
+++ b/scripts/name-a-register.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -e
+
+usage()
+{
+  echo -e "usage: ./name-a-register.sh phase register description\ne.g. ./name-a-register.sh alpha country \"Countries recognised by the UK\""
+}
+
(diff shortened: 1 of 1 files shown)


michaelabenyohai:
Should we change name-a-register.md as part of this?

nacnudus:
Good point @michaelabenyohai.  How about I delete `scripts/name-a-register.md` and add a section to `scripts/readme.md` in a separate PR?

michaelabenyohai:
@nacnudus sounds good to me!

PR #513, opened 2018-06-22
Add to scripts readme about naming registers

nacnudus:
### Context

There is now a script `scripts/name-a-register.sh` for setting/changing human-readable names of registers.

1. Delete the obsolete instructions in `scripts/name-a-register.md`
2. Add a section to `scripts/readme.md` for usage of the new script.


scripts/readme.md
```diff
@@ -92,7 +92,17 @@ Examples:
     Instances of ORJ serving the Field and Register Registers read files containing fields and registers YAML from S3 on startup.
 
     There is an Ansible task in the deployment repo to generate the files in S3 from the application. You will need to run this after reloading the metadata RSF.
-    
+
+### Naming a register with a human-friendly title
+
+The name of a register is immutable because it is its unique ID.  But the
+register can also have a human-readable name, or 'title'.  To set/change it, use
```
gidsg: I would rather call it `register-name` as this is what it is called in the RSF, rather than `title`.

Issue #514, opened 2018-06-29
Better way to load custom field descriptions


nacnudus:
### Situation

The current process to load custom field descriptions is clunky.

1. Commit custom descriptions to local master.
2. Load the register with `scripts/load-register-tsv.sh`, which defines the field in the RSF of the register and loads it together with the data.
3. Revert the local commit.

The custom descriptions aren't committed to GitHub unless by accident.  If they are committed, then they can be pushed to a register by accident (and this has happened in discovery/alpha).

Is one of the alternatives better?

#### Update the field descriptions later

1. Create the register RSF from master with default field descriptions
2. Update the register with custom descriptions using a script similar to [`scripts/name-a-register.sh`](https://github.com/openregister/deployment/blob/master/scripts/name-a-register.sh)

The script would obtain the latest item for the field from `GET /records/{key}/`, update it with a description passed at the command line, and post a new `item` and `entry` to the register.

The descriptions still wouldn't be pushed to GitHub by this method.

#### Commit to local branch

1. Commit everything except custom descriptions to the master branch
2. Start a new branch and commit the custom descriptions
3. Create the register RSF from the branch
4. Optionally push the branch to GitHub

The tests would have to be modified to allow a branch other than `master` to be used for constructing the register RSF.

The branch could be pushed to GitHub and left there, to record the custom descriptions for posterity (this is irrelevant for beta registers where the custom description will stay in the register).

This is more complicated than the current method.  Unless we left lots of branches hanging around, the custom descriptions wouldn't be kept in GitHub.

nacnudus:
pyyaml might support inheritance or "merging".  Search for `<:` here https://learnxinyminutes.com/docs/yaml/ (thanks @arnau)

arnau:
The README suggests there is already an approach to deal with register-specific scripts:

https://github.com/openregister/deployment#loading-data-to-registers-in-existing-phases-from-github

However, I'm failing to find such shellscript in the current master HEAD.

arnau:
So, I'm wondering if the specific field definitions should live in `[register]-data` repo as well?

arnau:
What sort of tests?

nacnudus:
https://github.com/openregister/registry-data/tree/master/tests

PR #516, opened 2018-07-03
Use custom field descriptions if available

nacnudus:
### Context

https://github.com/openregister/deployment/issues/514

When creating RSF with `scripts/rsfcreator.py` (called by several other scripts, in particular `scripts/load-register-tsv.sh`) look for a custom field definition in `registry-data/data/{phase}/{register}/field/{field}.yaml` and use it to override the definition in `registry-data/data/{phase}/field/{field}.yaml`.

A test is provided.

### Guidance to review

Check that it implements the first "alternative" discussed in the #514


nacnudus:
Now it overrides individual properties of fields, rather than the whole thing.

Not sure if `.__contains__()` is pythonic, what do you think @arnau?

scripts/rsfcreator.py
```diff
@@ -77,14 +75,20 @@ def rsf_for_line(line_dict, key_field, entry_type, key_prefix='', key=None):
             entry_type, key_prefix, key, timestamp, item_hash)
     return (item_line, entry_line)
 
+def merge_level_2(dct, other):
+    for k, v in other.items():
+        if (other.__contains__(k)):
```
arnau: The use of underscored methods should be avoided as much as possible. A common way to check the same would be:

```python
if k in other:
  dct[k] ...
```

scripts/rsfcreator.py
```diff
@@ -77,14 +75,20 @@ def rsf_for_line(line_dict, key_field, entry_type, key_prefix='', key=None):
...
+
 def generate_rsf(args):
     if args.register_data_root:
         register_def = read_register_from_local(args.phase, args.register_name, args.register_data_root)
         field_names = register_def['fields']
         fields_by_name = {fn: read_field_from_local(args.phase, fn, args.register_data_root) for fn in field_names}
         custom_fields_by_name = {fn: read_custom_field_from_local(args.phase, args.register_name, fn, args.register_data_root) for fn in field_names}
-        custom_fields_by_name = {k: v for k, v in custom_fields_by_name.items() if v}
-        fields_by_name = {**fields_by_name, **custom_fields_by_name}
+        custom_fields_by_name = remove_blanks(custom_fields_by_name)
```
arnau: On second thought, perhaps the two expressions that generate `custom_fields_by_name` could be extracted as a function so the logic is isolated (and can be tested as well), something on the lines of:

```python
def custom_fields(phase, register_name, register_data_root, field_names):
    fields = {}
    for name in field_names:
        field = read_custom_field_from_local(phase, register_name, name, register_data_root)
        if field is not None:
            fields[name] = field

    return fields

def generate_rsf(args):
  ...
  custom_fields_by_name = custom_fields(args.phase, args.register_name,  args.register_data_root, field_names)
```

nacnudus:
There's now `read_global_field_from_local()`, `read_custom_field_from_local()`, and a wrapper `read_field_from_local()` that merges global with custom when a custom one exists.

The deep merge function is no longer needed.

PR #517, opened 2018-07-06
New metadata

nacnudus:
### Context

Firebreak :)

The `register` register doesn't have fields for `custodian` or `register-name` (human-friendly name).  Hence there is nowhere to record that data in the register definition yaml files in `registry-data/data/{phase}/register/{register}.yaml`.

The `custodian` is added by a command-line argument to `scripts/rsfcreator.py`

The `register-name` is added by `scripts/name-a-register/sh`.

Both these steps have sometimes been forgotten when recreating a draft register, causing custodians to ask why the names have reverted to default values.

Alter `scripts/rsfcreator.py` to use extra metadata defined in `registry-data/data/{phase}/{register}/meta.yaml` if available.

Remove the `-- custodian` command-line flag from all scripts. 

### Guidance to review

Is this a sensible approach?  Currently there are no tests in the `registry-data` repo to check that the metadata yaml is sensible.

Check that the tests test the new behaviour, and still check the remaining behaviour.


michaelabenyohai:
Given that custom field descriptions live here:

`registry-data/data/{phase}/{register}/field/{field}.yaml`

Would it be better if metadata lived somewhere like:

`registry-data/data/{phase}/{register}/meta.yaml`

Then all register-specific stuff is available in one directory?

scripts/test-data/custom-field-description-meta.tsv
```diff
@@ -0,0 +1,3 @@
+custom-field-description	name	notifiable-locations	notifiable-animal-disease-investigation-category	notifiable-animal-disease-confirmation-category	start-date	end-date
```
michaelabenyohai: Am I being silly, I don't see where this file and the `custom-field-descriptions-meta.rsf` file are used?

nacnudus:
Thinking aloud about the file paths...

The metadata location `registry-data/data/{phase}/meta/{register}.yaml` follows the existing pattern `registry-data/data/{phase}/register/{register}.yaml`.  Both are register-specific, and the `meta` file is for things that we might have put in the register register.

If we're moving away from the register register concept then @michaelabenyohai's suggestion makes more sense `registry-data/data/{phase}/{register}/meta.yaml` and leaves room to migrate normal register metadata (name, description, etc.) to `registry-data/data/{phase}/{register}/register.yaml`

PR #519, opened 2018-07-13
Script to generate RSF to change register descriptions

MatMoore:
### Context
For https://trello.com/c/yljdvr9J/2635-bau-update-register-descriptions-for-beta-registers we had to update descriptions for a bunch of registers.

This isn't yet in the managing registers tool, so we have to manually create `add-item` and `append-entry` RSF to add a system entry to the register.

This was tedious and a bit error prone so I scripted the RSF generation part.

I'd be happy to make this more general purpose at a later date but for now I've just implemented what I needed to get the job done.

### DIFF
--- a/scripts/update_descriptions.py
+++ b/scripts/update_descriptions.py
@@ -1,0 +1,47 @@
+import requests
+import sys
+import hashlib
+import re
+import json
+import datetime
+
+def sha256(value):
+    m = hashlib.sha256()
+    m.update(value)
+    return m.hexdigest()
+
+if __name__ == '__main__':
+    if len(sys.argv) != 4:
+        print('Usage: ... <register> <old_description> <new_description>', file=sys.stderr)
+        sys.exit(1)
+    register = sys.argv[1]
+    old_desc = sys.argv[2]
+    new_desc = sys.argv[3]
+    print(f'<{register}> change "{old_desc}" -> "{new_desc}"', file=sys.stderr)
+
+    response = requests.get(f'https://{register}.beta.openregister.org/download-rsf')
+
+    old_item = None
+    old_hash = None
+    old_add_entry = None
+    for line in response.iter_lines():
+        match = re.match(rf'^add-item\t(.*{old_desc}.*)'.encode('utf8'), line)
+        if match:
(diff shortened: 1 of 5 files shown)


scripts/update_descriptions.py
```diff
@@ -0,0 +1,47 @@
...
+    new_desc = sys.argv[3]
+    print(f'<{register}> change "{old_desc}" -> "{new_desc}"', file=sys.stderr)
+
+    response = requests.get(f'https://{register}.beta.openregister.org/download-rsf')
+
+    old_item = None
+    old_hash = None
+    old_add_entry = None
+    for line in response.iter_lines():
+        match = re.match(rf'^add-item\t(.*{old_desc}.*)'.encode('utf8'), line)
```
michaelabenyohai: This works as long as we don't come across any other items with the same description in it. This is rare but sometimes we do have occasions where a register and an individual field will have the same description. This means that we risk picking up a `field:x` item here and accidentally combining it with a `register:x`entry.

Could you update the logic to scan for the correct item based on knowledge of its hash? So scan for the entry with key `register:x`, get the item-hash, then scan the items for the item with the corresponding hash. The item always occurs before the entry, which means either scanning the RSF file twice, or storing items and their item-hashes as you scan them. Not quite ideal for a little script but better than the risk of creating some dud RSF?

michaelabenyohai: Although using this approach you would have to make sure you pick up the latest entry with key `register:x` as it's possible for there to be multiple versions of it.

MatMoore: Ok, I'll have a go at this. It's probably safer to just parse all the item JSONs and look up the description key that way.

MatMoore: I had a go at implementing this yesterday afternoon. It's more complicated than the previous version but it's stricter. Is this what you had in mind? (see new commit)

MatMoore:
@michaelabenyohai I've rewritten the script to be stricter about identifying the existing item, and I've added some tests. Can you take another look?

gidsg:
Looks good to me, can you confirm @michaelabenyohai since you requested changes?

PR #520, opened 2018-07-27
Remove terrafrom code and config for influxcloud

michaelabenyohai:
### Context
We used to use InfluxCloud to store dropwizard metrics from openregister-java when it was hosted on AWS EC2. When we moved to PaaS we never reimplemented this feed in favour of using a Reliability Engineering provided solution. Although we don't have that yet, any work we do to reimplement metrics from ORJ should be in conjunction with Reliability Engineering's tool that we use for our Ruby application. We therefore no longer need this code.

We can then stop paying for InfluxCloud subscription.

Remove all influxcloud code. This is not actually being used anywhere.

### Guidance to review
Run `make plan -e vpc=beta` and you should see the deletion of `telegraf.conf` from S3 bucket containing beta config files. We used to use telegraf as part of our AWS solution but we no longer use it and this file isn't needed.


influxdb/influxdb_configuration.txt.tmpl
```diff
@@ -1,13 +0,0 @@
-# DDL
-CREATE DATABASE metrics
```
gidsg: where does this database live? Is it something we need to delete?

aws/registers/Makefile
```diff
@@ -25,9 +25,6 @@ export TF_VAR_pingdom_user:=$(or $(shell PASSWORD_STORE_DIR=~/.registers-pass pa
 export TF_VAR_pingdom_password:=$(or $(shell PASSWORD_STORE_DIR=~/.registers-pass pass services/pingdom | head -n 1), $(error "Couldn't get Pingdom password"))
 export TF_VAR_pingdom_api_key:=$(or $(shell PASSWORD_STORE_DIR=~/.registers-pass pass services/pingdom/api | head -n 1), $(error "Couldn't get Pingdom API key"))
 
-# InfluxCloud
-export TF_VAR_influxdb_password:=$(or $(shell PASSWORD_STORE_DIR=~/.registers-pass pass services/influxcloud/telegraf | tail -n 1), $(error "Couldn't get InfluxCloud password"))
```
gidsg: We should make a follow-up task to delete these credentials from `registers-pass`

influxdb/influxdb_configuration.txt.tmpl
michaelabenyohai: The database lives inside influxcloud so it should just be a case of cancelling our description

aws/registers/Makefile
michaelabenyohai: will do!

PR #523, opened 2018-08-15
Remove letsencrypt

michaelabenyohai:
### Context
We no longer use LetsEncrypt to create SSL certificates for *.register.gov.uk domains so we can remove the terraform and ansible code required to do this. We now use AWS Certificate Manager, where certificate renewal and validation is automated.

- Remove `ansible/letsencrypt.yml`, the script for generating an SSL certificate
- Remove `aws/tls-certs` directory, the terraform code for uploading the SSL certificate to AWS IAM
- Remove all steps from relevant READMEs

### Guidance to review


michaelabenyohai:
Certificate Manager should \*\*just work\*\* now that the certificates have been created. I.e. they should just renew themselves without any manual steps. This is because they use DNS validation - as long as the DNS records are never deleted from Route 53 validation (and therefore renewal) should pass automatically.

The README currently only contains documentation on how to make new registers and so I'm not sure documenting how ACM works really fits here. But let me know if you think otherwise? I think there's probably docs missing on how our infrastructure fits together and maybe that's something we could create and add this to? What do you think @MatMoore?

PR #525, opened 2018-08-24
CloudFront Invalidation Lambdas 

gidsg:
### Context
Reduce latency of update visibility on CDN 

The approach is similar to that taken for tracking API Keys:
* Lamda@Edge parses Origin Request POSTs and writes metadata to CloudWatch Logs
* A second Lambda consumes these logs and triggers a CloudFront Invalidation Request against a list of paths

### Guidance to review
The Lambdas are deployed on `*.test.openregister.org` so you can test the flow by POSTing to `load-rsf` in this environment.
Note that a limitation of Origin Requests is that they have not been processed, so we trigger an invalidation request even if the POST is not accepted by ORJ. (This is a trade-off as intercepting at this point in the request-response cycle allows us to access the POST body which may be useful data for downstream consumers.

Paired with @MatMoore 

-----------

Todo 
- [x]  terraform cache-invalidator lambda
- [x]  add cloudwatch alarm for lambda failures
- [x] check what Cache Behaviour value should be for Lambda
- [x] make decision about POST host
- [ ] add tests for lambdas
- [x] use terraform archive_file to perform zip to resolve source_code_hash issue https://github.com/hashicorp/terraform/issues/14395
- [x] match beta.openregister.org to register.gov.uk distribution


aws/lambda/lambda-edge-lambdas.tf
```diff
@@ -77,7 +76,22 @@ resource "aws_lambda_function" "api-key-to-cloudfront-logs" {
...
   publish          = true
 }
+resource "aws_lambda_function" "cloudfront-post-logger" {
+  function_name    = "cloudfront-post-logger"
+  filename         = "build/node/cloudfront-post-logger.zip"
+  source_code_hash = "${base64sha256(file("build/node/cloudfront-post-logger.zip"))}"
+  handler          = "lambda.handler"
+  role             = "${aws_iam_role.lambda-edge-role.arn}"
+  memory_size      = "128"
+  runtime          = "nodejs6.10"
```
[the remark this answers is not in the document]
gidsg: There is none, it was just because I used existing boilerplate. Have bumped version now!

aws/lambda/node/README.md
```diff
@@ -0,0 +1,21 @@
+# Node Lambdas
+
+## Running Locally
+
+### Setup
+Install [aws-sam-cli](https://github.com/awslabs/aws-sam-cli/blob/develop/docs/installation.rst)
```
arnau: Would it be better to use a link to either `master` or a specific version instead of `develop`?

gidsg:
I think this is ready for review now, there are some caveats I have learned whilst Terraforming:
* The `IncludeBody` boolean is not supported by Terraform. As we are not using the body currently I have just commented out the usage for now
* Terraform is unable to create email subscriptions to the alarms. Therefore we will need to manually subscribe the SNS to our email address
* I have not added tests. We have the ability to run the lambdas locally using SAM local. I think this is good enough for a first-pass. Ideally we would add unit tests and a linter to run in Travis CI but as this looks non-trivial I would rather make a separate card to look into it.

aws/lambda/node/README.md
```diff
@@ -0,0 +1,21 @@
+# Node Lambdas
+
+## Running Locally
+
+### Setup
+Install [aws-sam-cli](https://github.com/awslabs/aws-sam-cli/blob/master/docs/installation.rst)
+
+### Running lambdas
+From `deployment/aws/node` directory:
```
MatMoore: this should be aws/lambda/node.

aws/lambda/node/README.md
```diff
@@ -0,0 +1,21 @@
+# Node Lambdas
+
+## Running Locally
+
+### Setup
+Install [aws-sam-cli](https://github.com/awslabs/aws-sam-cli/blob/master/docs/installation.rst)
```
MatMoore: this documentation makes it seem like you have to mess around with paths, but if you pip install it into a virtualenv it just works. So we could add it to our requirements.txt in scripts.

aws/lambda/node/cache-invalidator/lambda.js
```diff
@@ -0,0 +1,69 @@
...
+  const messageObject = messageParts[2];
+  const parsedMessageObject = JSON.parse(messageObject);
+  console.log(parsedMessageObject);
+  const {distributionId} = parsedMessageObject;
+  const callerReference = 'cache-invalidator-lambda-' + new Date().toISOString();
+  const {host} = parsedMessageObject;
+  const isBeta = host.split('.')[1] === 'beta';
+
+  const getDistributionId = new Promise((resolve, reject) => {
+    if (isBeta) {
```
MatMoore: I think you could also write this without another promise by using await on the inner promise.

Something like this

```javascript
if (isBeta) {
   try {
     const data = await cloudfront.listDistributions({MaxItems: '1000'}).promise()
     // set getDistributionId from data
   } catch err {
   }
} else {
  getDistributionId = distributionId;
}
```

not 100% sure about this though, as I'm new to this syntax

PR #528, opened 2018-08-30
use terraform archive_file to generate node lambda zip files

gidsg:
### Context
Previously we generated zip files in the Makefile, this caused a bug whereby the `source_code_hash` would change even if the code had not changed. See https://github.com/hashicorp/terraform/issues/14395

* Use terraform provided [archive_file](https://www.terraform.io/docs/providers/archive/d/archive_file.html) to generate zips. This is the route recommended in: https://github.com/hashicorp/terraform/issues/14395#issuecomment-300794749
* Remove redundant build steps for node lambdas. As they were not doing any transformation in the Makefile, they serve no purpose.
* The node lambdas do not have any external npm dependencies so we just zip the single file `lambda.js`. This makes the `source_code_hash` consistency more reliable.
* Removes the possibility of injecting secrets then renaming. This should no longer be required as we should use [Lambda Environment Variables](https://docs.aws.amazon.com/lambda/latest/dg/env_variables.html) for any secrets

### Guidance to review
`make plan` should show change in `source_code_hash`. This is a refactor, there should be no change in functionality after applying.


aws/lambda/Makefile
```diff
@@ -46,7 +46,7 @@ destroy: configure-state
 build/node/%.zip: node/%/lambda.js node/%/package.json
```
MatMoore: I think here we should change the target of this step to be the directory, as it no longer produces the zip. Otherwise this could execute/not execute in the wrong circumstances.

aws/lambda/Makefile
```diff
@@ -46,7 +46,7 @@ destroy: configure-state
...
 	mkdir -p $(dir $@)
 	npm install --production --prefix $(dir $<)
-	cd $(dir $<); zip -r ../../$@ $(notdir $<) node_modules/
```
MatMoore: this whole line isn't needed

aws/lambda/Makefile
gidsg: I think this build step is still needed to copy a subset of the files to another folder?

PR #529, opened 2018-09-06
Extend scripts for updating register metadata

MatMoore:
### Context
For https://trello.com/c/J1j2gREF/2809-update-register-user-friendly-titles-and-descriptions I needed to update friendly titles and descriptions for a bunch of registers.

These pieces of metadata are stored in the system entries of the register with these keys:
- `register:{register name}`
- `register-name`

We are also storing copies of this metadata in github. We discussed whether we wanted to keep doing this, and the decision was to keep these two data sources in sync until we have a proper backup strategy for registers.

I previously created some python code for modifying the `register:{register name}` item to update a register's description (https://github.com/openregister/deployment/pull/519)

Duncan added the name-a-register script in https://github.com/openregister/deployment/pull/512 which sets `register-name`.

- I've factored out my python code for doing stuff with RSF into a standalone package, rsflib, so it's easier to reuse. For some tasks it's relatively easy to craft some RSF, but when you're modifying an existing item it's much more error prone, so having a package makes it easier.

- I've created a new script to update both kinds of metadata at the same time (we don't have to commit this: it's slightly more useful than running two commands for every register, but not as user friendly as it could be)

- I've added some more info to the readme about the register-data repository

- I've written a script that compares the github repository to the registers, so we can see whether it's consistent. At the moment, everything I've updated is out of sync with github, but somebody is going to update github to fix this.

### Guidance to review
- Are we happy with this approach for now?
- Is there anything I should take out of this PR?
- Is the information in the readme correct?


scripts/readme.md
```diff
@@ -110,20 +123,38 @@ Run this using:
...
 
 ```
-./rsf-load.sh "https://${register}.beta.openregister.org" openregister `register-pass beta/app/mint/$register` < ${register}_update.rsf
+python update_descriptions.py $register $desc | ./rsf-load.sh "https://${register}.beta.openregister.org" openregister `registers-pass beta/app/mint/$register` < ${register}_update.rsf
 ```
 
 This example assumes that the register is in beta.
 
-This will update the API explorer, but at the time of writing, registers frontend won't automatically pick up the description, because system entries are not included in incremental updates.
+This will update the API explorer, but at the time of writing, registers frontend won't automatically pick up the description, because system entries are not included in incremental updates. To do this, follow the [instructions for redownloading a register in registers frontend](https://github.com/openregister/registers-frontend#populating-the-database-with-register-data-on-paas).
```
gidsg: If you want, this script could also perform the re-index after updating e.g.
```
cf run-task registers "bundle exec rake registers_frontend:populate_db:force_full_register_download[$register]" -m 1G --name force_full_reindex
```
we would need to make sure you were logged in and in the correct space though.

MatMoore: I think I'll leave this bit out for now, because it depends on our local environments being set up in a particular way, and I'd rather run this part one-by-one anyway.

This time round I messed up the frontend by running this, so I think it makes sense to test populating the registers locally before doing it in production.

MatMoore:
@gidsg I think this is ready for review now.

scripts/readme.md
```diff
@@ -18,6 +18,18 @@ where $PHASE is test, discovery, alpha or beta
 
 Within those directories the definitions of individual fields or registers are in separate YAML files.
 
+This repo also contains files like
+
+    $ROOT/openregister/registry-data/data/$PHASE/$REGISTER/meta.yaml
+
+These contain additional metadata about the register. Each key corresponds to system entries stored in the register, for example "friendly-name".
```
gidsg: Did you mean `register-name` rather than `friendly-name`?

scripts/readme.md
```diff
@@ -18,6 +18,18 @@ where $PHASE is test, discovery, alpha or beta
...
+
+We intend to keep this github repository in sync with the actual registers until we have a way to archive register data. You can check that the data matches up by running
+
+    OAUTH_TOKEN=<github personal access token> python validate_metadata.py
```
gidsg: can we make this `GITHUB_OAUTH_TOKEN` to be more specific?

gidsg: can we specify `python3`? As the script did not work for me using `python`.

scripts/readme.md
```diff
@@ -110,20 +122,38 @@ Run this using:
...
 ```
 
-Then load the RSF using:
+Then check that the RSF looks correct.
+
+To actually load the RSF, pipe the output to `rsf-load.sh`:
 
 ```
-./rsf-load.sh "https://${register}.beta.openregister.org" openregister `register-pass beta/app/mint/$register` < ${register}_update.rsf
+python update_descriptions.py $register $desc | ./rsf-load.sh "https://${register}.beta.openregister.org" openregister `registers-pass beta/app/mint/$register` < ${register}_update.rsf
```
gidsg: can we specify `python3`?

scripts/readme.md
```diff
@@ -110,20 +122,38 @@ Run this using:
...
+
+You can change the name and description of a register in one step:
+
+```
+register=country
+name="Country register"
+desc="British English names of all countries currently recognised by the UK government"
+
+# Preview the generated RSF
+python update_titles_and_descriptions.py "$register" "$name" "$desc"
```
gidsg: can we specify `python3` here? I think the script would be better called `update_name_and_description.py` as title is not how we refer to it in RSF, so better to keep things consistent.

scripts/readme.md
MatMoore: oops, yes I did

scripts/rsflib/parser.py
```diff
@@ -0,0 +1,103 @@
...
+
+class RsfFile:
+    """
+    A parsed RSF file
+    """
+    def __init__(self, register_name, lines):
+        """
+        Parse an RSF file into commands.
+        register_name is the internal ID of the register.
+        lines should contain binary strings.
```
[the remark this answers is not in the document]
MatMoore: I guess I meant bytestrings here. Python has two string types: unicode and bytes.

scripts/readme.md
MatMoore: Yep

MatMoore: Yep. The prerequisites mentions python3 but I'll update this to be consistent with the other examples

PR #533, opened 2018-09-14
Lambda to clean IP address from logs

MatMoore:
### Context
We keep cloudfront logs in S3 for a period of time, but then we have to delete them since IP addresses are personal data.

As part of https://trello.com/c/J9eMZntG/2744-anonymise-or-remove-cloudfront-logs-in-s3 I'm looking at automating the removal of personal data from these logs so we can continue to provide historical data to [performance platform](https://www.gov.uk/performance/govuk-registers). 

This is a lambda function that should run every time an object is added to an S3 bucket. It cleans the data and then puts the object in another bucket. The idea is then to put a retention policy on the source bucket.

### Guidance to review
At this stage it would be useful to sanity check that this approach makes sense

- I've only tested the data cleaning part so far
- The entire file needs to be loaded into memory to gzip decode it. I think this is ok because cloudfront seems to split things into small files, but I don't know if this is always the case
- I haven't thought about how to monitor errors in this script yet


aws/lambda/python/log-anonymiser/lambda.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return gzip.compress(output)
+
+
+def lambda_handler(event, context):
+    record = event['Records'][0]
+    record_s3 = record['s3']
+    source_bucket = record_s3['bucket']['name']
+    source_key = unquote_plus(record_s3['object']['key'])
+    target_bucket = 'mat-test-b'
```
gidsg: We should use an environment variable for the target bucket name.

MatMoore:
The lambda itself is working now. I'm going to try terraforming it tomorrow. For now it only does the registers.gov.uk bucket. I don't think we actually need to retain openregister.org logs but will double check this before adding the retention policy.

aws/lambda/python/log-anonymiser/module.tf
```diff
@@ -0,0 +1,42 @@
+data "aws_iam_role" "lambda_role" {
+  role_name = "lambda"
```
gidsg: can we terraform this role, or use a terraformed existing role rather than `data`? I know we use this in some places but only as a workaround where a terraform defined role is not possible.

MatMoore: I think there's no reason not to, but the existing lambdas are using the "lambda" role. It would be good to chat about how best to do this and maybe pair on it later this week?

aws/lambda/python/log-anonymiser/module.tf
```diff
@@ -0,0 +1,123 @@
...
+        "arn:aws:s3:::cloudfront-logs-register-gov-uk/*"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:PutObject"
+      ],
+      "Resource": [
+        "arn:aws:s3:::cloudfront-logs-openregister-anonymised",
```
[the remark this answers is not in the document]
MatMoore: The reason I didn't do this was because the source bucket isn't terraformed and we decided not to terraform that. Since the source bucket would also have to be created manually if we set everything up from scratch, I drew the line between s3 and lambda so only everything in lambda is terraformed.

I'd also prefer for both buckets to be managed the same way as the settings should be consistent across both them.

aws/lambda/main.tf
```diff
@@ -104,6 +104,11 @@ module "cache_invalidator_eu_central_1" {
   region = "eu-central-1"
 }
 
+module "log-anonymiser" {
+  source = "python/log-anonymiser"
+  region = "eu-west-1"
```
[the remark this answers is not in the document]
MatMoore: Omitting this gives me the error:

```
* module root: module log-anonymiser: required variable "region" not set
make: *** [plan] Error 1
```

PR #534, opened 2018-09-19
Pass user agent from Lambda -> Google Analytics

MatMoore:
### Context
Currently every request that doesn't come from the GDS client library is marked as "API"

This masks the fact that traffic could be coming from browsers, bots (like pingdom), or non-gds client libraries.

This passes the value along so we can filter on it.

https://trello.com/c/aXdyrWK2/2901-add-user-agent-to-lambda-event-as-custom-dimension

### Guidance to review
<strike>I'm not actually using a custom dimension because GA has a user agent field that you can override.</strike>. The user agent is now stored in custom dimension 6.

The first lambda should log the user agent to cloudwatch (test with `sam local invoke CloudfrontLogAPIKeyFunction -e log-api-key-to-cloudwatch/example.json`)

The second lambda should read this value and include it in the call to GA.

`sam local invoke StoreAPIUsageFunction -e cloudfront-logs-api-key-to-google-analytics/examples/logs-no-key.json`


gidsg:
can we update the test events to reflect these fields?

aws/lambda/node/template.yml
```diff
@@ -25,3 +25,11 @@ Resources:
...
             Handler: lambda.handler
             Runtime: nodejs8.10
+
+
+    CloudfrontLogAPIKeyFunction:
+        Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+        Properties:
+            CodeUri: log-api-key-to-cloudwatch/
+            Handler: lambda.handler
+            Runtime: nodejs8.10
```
gidsg: I think runtime should be `6.10` to match what we currently have deployed.

aws/lambda/node/log-api-key-to-cloudwatch/lambda.js
```diff
@@ -32,11 +32,13 @@ exports.handler = (event, context, callback) => {
 	const hittype = request.headers['user-agent'] && request.headers['user-agent'][0].value == 'CL' ? 'Client Library' : 'API';
 	const queryParams = (request.querystring === '') ? '' : '?' + request.querystring;
 	const endpoint = 'https://' + host + request.uri + queryParams;
+	const useragent = request.headers['user-agent'][0].value;
```
gidsg: if we declare this variable further up, we can re-use it in `hittype`

aws/lambda/node/template.yml
```diff
@@ -25,3 +25,18 @@ Resources:
             CodeUri: cloudfront-post-logger/
             Handler: lambda.handler
             Runtime: nodejs8.10
+
+
+    CloudfrontLogAPIKeyFunction:
+        Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
```
gidsg: Can we add these with examples to the README for consistency?

aws/lambda/node/log-api-key-to-cloudwatch/lambda.js
```diff
@@ -29,14 +29,16 @@ exports.handler = (event, context, callback) => {
 
 	const apikey = request.headers.authorization ? request.headers.authorization[0].value : 'Anonymous';
 	const host = request.headers.host[0].value;
-	const hittype = request.headers['user-agent'] && request.headers['user-agent'][0].value == 'CL' ? 'Client Library' : 'API';
+	const useragent = request.headers['user-agent'] && request.headers['user-agent'][0].value;
```
gidsg: can we trim this to the maximum permitted characters?

MatMoore: I trimmed this in the lambda where it's passed to GA, so we still log out the full  thing.

MatMoore:
I'm pausing this while I'm on holiday, but if anyone wants to try deploying it please feel free to.

PR #538, opened 2018-10-09
redirect to remove trailing slash

gidsg:
### Context
Implement https://github.com/openregister/openregister-java/blob/master/doc/arch/adr-009-trailing-slash.md

Implement redirect in Lambda@Edge.

### Guidance to review
See test commands in README. I have also deployed this version of the Lambda to https://country.test.openregister.org


MatMoore:
What's the reason for not doing this in ORJ?

I think it makes sense for this to be in the app itself as it's not really adding anything on top of the standard API, it's just an implementation choice. It feels different to API keys, cache invalidation etc.

gidsg:
I did try and do this in ORJ initially but got into a fight with JAX-RS which wants to resolve both styles and I couldn't make it do the desired behaviour. See https://github.com/eclipse-ee4j/jaxrs-api/issues/468 also https://stackoverflow.com/a/15215730/238230

gidsg:
...actually I will try again to do it in ORJ, I have a different idea, let me try that before closing this PR.

gidsg:
closing in favour of https://github.com/openregister/openregister-java/pull/503

PR #546, opened 2018-11-08
add script for deleting a register

gidsg:
### Context
Make it easier to delete data from non-beta registers

Add a script to delete data from non-beta registers

### Guidance to review
Should work as per documentation (hard to test end-to-end though without actually deleting a register!)


trvllngmn:
I've run the script and it has performed as expected

PR #547, opened 2018-11-08
add update custodian script

gidsg:
### Context
I needed this recently so worth adding to the repo

Adds update custodian script (based on `name-a-register.sh` script)

### Guidance to review
We could generecise common code from `name-a-register.sh` and `update-custodian.sh` into `register-actions.sh` as something like `update-system-key` but I wasn't sure it was worth the effort v.s. copy paste. Do you have strong feelings about this @MatMoore?


scripts/update-custodian.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -e
```
MatMoore: Can we add -u and -o pipefail to this? http://blog.kablamo.org/2015/11/08/bash-tricks-eux/

scripts/update-custodian.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "CUSTODIAN - $CUSTODIAN"
+
+OPENREGISTER_BASE="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
+echo "OPENREGISTER_BASE - $OPENREGISTER_BASE"
+source "$OPENREGISTER_BASE/deployment/scripts/includes/register-actions.sh"
+
+TIME=`date +%Y-%m-%dT%H:%M:%SZ`
+JSON="{\"custodian\":\"$CUSTODIAN\"}"
+SHA256=`echo -n $JSON | shasum -a 256 | head -c 64`
+PASSWORD=`PASSWORD_STORE_DIR=~/.registers-pass pass $PHASE/app/mint/$REGISTER`
```
arnau: Does this override any env variable you might have in your environment? If so, should we respect the environment?

gidsg: fixed in 2ccaa815b41e24935c7f35d60f8f9f9db7c18c77

PR #562, opened 2019-02-08
Pingdom updates and clinical commissioning group promotion to beta

oswaldquek:


gidsg:
Just for context the terraform pingdom provider latest version has removed these fields as they are deprecated in the Pingdom API.

PR #574, opened 2019-03-28
Rewrite upload scripts

arnau:

This PR unifies the update scripts under a single CLI umbrella. Also, removes shellscript and uses python 3 for all operations.

It is part of the same effort as this PR https://github.com/openregister/registry-data/pull/293

### Guidance to review

The readme in ./scripts/README.md should be self explanatory.

* All update operations work as before (update data from tsv, change custodian, register title and register description and field description).


scripts/README.md
```diff
@@ -0,0 +1,177 @@
+# Deployment scripts
+
+This directory contains two types of script: [operate test and discovery
+environments](../create-a-discovery-register-environment.md) and upload data
+to the Open Register Java (ORJ) application.
+
+
+## Introduction
+
+The raw data for every GDS-hosted register is in Github:
```
gidsg: this isn't the case currently. I don't think we should say this in the README until it reflects the current reality.

scripts/README.md
```diff
@@ -0,0 +1,177 @@
...
+help:
+
+```
+pipenv run update --help
+```
+
+
+## Update operations
+
+Note that in order to get any metadata change appear on the frontend you will
```
gidsg: should be _to appear_

scripts/README.md
arnau: Once this PR is merged it will be true (see linked PR)

scripts/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+## The Field and Register registers
+
+
+Instances of ORJ serving the Field and Register Registers read yaml files
+containing fields and registers from S3 on startup.
+
+There is an Ansible task in the deployment repo to generate the files in S3
+from the application. You will need to run this after reloading the metadata
+RSF.
```
gidsg: it is not only the ansible task as you also need to re-deploy the application. Better to link to: https://github.com/openregister/deployment/blob/master/docs/extra-steps-beta.md#4-run-ansible-task steps 4 and 5.

scripts/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+The update CLI is in the `update` python package in this directory. It is
+defined as a Pipenv script in `Pipenv`.
+
+The logic around serialising/deserialising RSF and handling patches and
+registers is in the `registers` python package part of the Registers CLI (See
+https://github.com/openregister/registers-cli).
+
+
+If you need to change how the update CLI handles pass, orj or the CLI itself,
```
gidsg: I would put at least `pass` in backticks so as not to confuse the language meaning and the application name. Also should be ORJ in capitals or write out the whole repo name, to reduce potential confusion.

scripts/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Setup
+
+### Dependencies
+
+Before attempting to use any update operation, please ensure all dependencies
+are correctly installed and set up in your machine.
+
+* [Registers Credentials set up](https://github.com/openregister/credentials).
+* Python 3.
```
gidsg: should we specify Python 3.7?

scripts/README.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Check the context subcommand help for more details:
+
+```sh
+pipenv run update context --help
+```
+
+
+### Change a field description
```
gidsg: Might be good to clarify heading to Changing a field description _for a single register_? As non overridden descriptions are derived from the Field register.

scripts/README.md
arnau: Pipenv locks the version to 3.7 already (one of the reasons for using Pipenv).

PR #575, opened 2019-04-15
document ORJ deployment process

gidsg:
### Context
process was not previously fully documented

document ORJ deployment process

### Guidance to review
Check documentation makes sense

### DIFF
--- a/docs/openregister-java-deployment.md
+++ b/docs/openregister-java-deployment.md
@@ -1,0 +1,23 @@
+# openregister-java deployment process
+
+## Deploying
+* Merge to `master`
+*Tests and building of JAR takes place in Travis-CI*
+* Build is automatically deployed to `test` environment e.g. https://country.test.openregister.org/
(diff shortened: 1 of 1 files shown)


docs/openregister-java-deployment.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Deployment Notifications
+Deployment notifications are sent to Slack using a webhook integration.
+
+### Changing the Notification configuration
+The architecture of the notifier is:
+```
+CodePipeline --> CloudWatch Logs --> CloudWatch Rules --> SNS --> Lambda --> Slack
+```
+Note: The notification architecture is not currently Terraformed. The Lambda is `deployment_to_slack` in `eu-west-1`.
```
phantas: how is the token for the Slack webhook managed? For example, what would I need to do to update the token?

phantas: And is the token only used for this integration?

